PR #163, opened 2019-12-19
validate pub/sub operations on server

danrubel:
This updates the client to include the current user's firebase user-id-token in the pub/sub request and updates the firebase function to validate that token before publishing the message.

publish log with success

In addition:
1) PubSubClient functionality has been extracted into a class as a move towards injecting both firebase and pub/sub at the app level.
2) firebase function has been improved to log error/stacktrace if the required request fields are missing.

publish log with error

### DIFF
--- a/webapp/lib/platform.dart
+++ b/webapp/lib/platform.dart
@@ -19,7 +19,6 @@ PubSubClient _pubsubInstance;
 
 init() async {
   await platform_constants.init();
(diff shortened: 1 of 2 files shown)


lukechurch:
HI @danrubel - Oops, very sorry I missed your comment on this. When you do something like this, please can you re-request a review (circle icon next to my name). Otherwise it doesn't get pushed back into the queue of reviews pending my attention.

PR #165, opened 2019-12-30
add Message.status enum to track progress through rapidpro

danrubel:
This adds a new MessageStatus enum field in each message to track messages that are sent from the nook client through rapid_pro. 
* `pending` - message was sent to rapid_pro but send has not yet been confirmed
* `confirmed` - confirmation received from rapid_pro that message was sent
* `failed` - message send failed somewhere in the process
* `unknown` - catch all for non-of-the-above until all messages have a status


danrubel:
Work towards completing https://github.com/larksystems/Project-2019-WorldBank-PLR/issues/28

PR #166, opened 2019-12-30
system messages

danrubel:
This adds the capability for displaying admin-level messages across all clients via a new `systemMessages` collection in firebase... but does not actually display those messages yet.


lukechurch:
LGTM - @marianamarasoiu can you take a look and see if this looks like it'll be easy to integrate with UI elements?

PR #171, opened 2020-01-01
Add snackbar notification UI

marianamarasoiu:
Hi @lukechurch and @danrubel 

As discussed, this adds a notification via a snackbar at the bottom of the screen similar to the one from Coda. Let me know if you have any feedback on the design - happy to make any adjustments etc.

Screenshot:
Screenshot 2020-01-01 at 21 27 46

Thanks!

### DIFF
--- a/webapp/lib/view.dart
+++ b/webapp/lib/view.dart
@@ -34,7 +34,7 @@ void init() {
   querySelector('header')
     ..append(authHeaderView.authElement);
 
-  document.onKeyPress.listen((event) => command(UIAction.keyPressed, new KeyPressData(event.key)));
+  document.onKeyDown.listen((event) => command(UIAction.keyPressed, new KeyPressData(event.key)));
 }
 
 void initSignedInView() {
@@ -1056,7 +1056,7 @@ class SnackbarView {
   }
 
   hideSnackbar() {
-    snackbarElement.classes.add('hidden');
+    snackbarElement.classes.toggle('hidden', true);
     snackbarElement.attributes.remove('type');
     // Remove the contents after the animation ends - the duration here must match the animation lenth
     new Timer(new Duration(milliseconds: 200), () => _contents.text = '');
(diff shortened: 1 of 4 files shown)


webapp/lib/view.dart
```diff
@@ -1019,3 +1024,41 @@ class UrlView {
...
+    _contents.text = message;
+    snackbarElement.classes.remove('hidden');
+    snackbarElement.setAttribute('type', type.toString().replaceAll('SnackbarNotificationType.', ''));
+    new Timer(new Duration(seconds: 3), () => hideSnackbar());
+  }
+
+  hideSnackbar() {
+    snackbarElement.classes.add('hidden');
+    snackbarElement.attributes.remove('type');
+    // Remove the contents after the animation ends - the duration here must match the animation lenth
```
lukechurch: nit: s/lenth/length

lukechurch: Could you make it explicit that the thing it has to match is defined in snackbar.css?

webapp/lib/view.dart
```diff
@@ -1019,3 +1024,41 @@ class UrlView {
...
+    _contents = new DivElement()
+      ..classes.add('contents');
+    snackbarElement.append(_contents);
+  }
+
+  showSnackbar(String message, SnackbarNotificationType type) {
+    _contents.text = message;
+    snackbarElement.classes.remove('hidden');
+    snackbarElement.setAttribute('type', type.toString().replaceAll('SnackbarNotificationType.', ''));
+    new Timer(new Duration(seconds: 3), () => hideSnackbar());
```
lukechurch: Could you a couple of these magic numbers into named constants defined on the class? E.g. I think this is the amount of time that the snackbar will be visible for?

webapp/lib/view.dart
marianamarasoiu: Done, thanks!

webapp/lib/view.dart
marianamarasoiu: Good idea, done now!

marianamarasoiu:
Thanks both for the reviews, merging now!

PR #178, opened 2020-01-14
Refactor model use of firebase

danrubel:
Currently, the nook model is tightly coupled to firebase, which prevents local testing and driving the model using json or other storage solutions. This PR is the first step in refactoring the model so that firebase is not directly used by the model.

See https://github.com/larksystems/Project-2019-WorldBank-PLR/issues/31

### DIFF
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -14,7 +14,7 @@ class Conversation {
   String notes;
   bool unread;
 
-  static Conversation fromFirestore(DocSnapshot doc, [Conversation modelObj]) =>
+  static Conversation fromSnapshot(DocSnapshot doc, [Conversation modelObj]) =>
       fromData(doc.data, modelObj);
 
   static Conversation fromData(data, [Conversation modelObj]) {
@@ -28,7 +28,7 @@ class Conversation {
   }
 
   static void listen(firestore.Firestore fs, ConversationCollectionListener listener, String collectionRoot) =>
-      listenForUpdates<Conversation>(fs, listener, collectionRoot, Conversation.fromFirestore);
+      listenForUpdates<Conversation>(fs, listener, collectionRoot, Conversation.fromSnapshot);
 
   Map<String, dynamic> toData() {
     return {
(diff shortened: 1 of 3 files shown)


lukechurch:
Is DocSnapshot defined? is it possible you need to check in an update to the model as well? If I'm wrong, PLMK and I'll pull all of the local code and understand how it's working rather than just reading this PR.

danrubel:
Yes, `DocSnapshot` is part of this PR can be see at the [bottom of the file in the first commit](https://github.com/larksystems/nook/pull/178/commits/eb7f5b81fa6b2bc674445979315c8eb20859a9e9)

danrubel:
@lukechurch Yes, `DocSnapshot` is part of this PR can be see at the [bottom of the file in the first commit](https://github.com/larksystems/nook/pull/178/commits/eb7f5b81fa6b2bc674445979315c8eb20859a9e9)

danrubel:
@lukechurch I merged a stacked PR into this PR before noticing that you had not approved this PR. Sorry about that.

webapp/lib/model.g.dart
```diff
@@ -411,11 +411,49 @@ void listenForUpdates<T>(
...
 
+/// A snapshot of a document's id and data at a particular moment in time.
+class DocSnapshot {
+  final String id;
+  final Map<String, dynamic> data;
+
+  DocSnapshot(this.id, this.data);
+}
+
+/// A batch update, used to perform multiple writes as a single atomic unit.
```
lukechurch: Comments as elsewhere about these being a batch not an atomic unit.

webapp/lib/platform.dart
```diff
@@ -138,13 +141,14 @@ Future updateUnread(List<Conversation> conversations, bool newValue) async {
...
   int batchSize = 0;
   for (var conversation in conversations) {
     conversation.updateUnread(_firestoreInstance, conversation.documentPath, newValue, batch);
+    ++batchSize;
     if (batchSize == _MAX_BATCH_SIZE) {
       await batch.commit();
-      batch = _firestoreInstance.batch();
+      batch = FirestoreDocUpdateBatch(_firestoreInstance.batch());
       batchSize = 0;
     }
```
lukechurch: I think we also need to commit the final set after the user interaction - the problem is things like what happenes if we do a bunch of stuff and then close the window. We have to commit all of the outstanding batched before the application closes. 

Given typical write performance on Firebase, I think we can be pretty eager about calling commit

danrubel: Done.

webapp/lib/model.g.dart
danrubel: I think that as of the latest update, they are now atomic.

PR #179, opened 2020-01-16
replace firestore.WriteBatch with DocUpdate

danrubel:
As another step towards [removing firebase from the model](https://github.com/larksystems/Project-2019-WorldBank-PLR/issues/31), this PR replaces `firestore.WriteBatch` with `DocUpdate`.

### DIFF
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -42,28 +42,28 @@ class Conversation {
 
   DocUpdate updateTagIds(firestore.Firestore fs, String documentPath, List<String> newValue, [DocUpdate batch]) {
     tagIds = newValue;
-    batch ??= FsDocUpdate(fs.batch());
+    batch ??= FirestoreDocUpdate(fs.batch());
     batch.update(fs.doc(documentPath), data: {'tags': newValue});
     return batch;
   }
 
   DocUpdate updateMessages(firestore.Firestore fs, String documentPath, List<Message> newValue, [DocUpdate batch]) {
     messages = newValue;
-    batch ??= FsDocUpdate(fs.batch());
+    batch ??= FirestoreDocUpdate(fs.batch());
     batch.update(fs.doc(documentPath), data: {'messages': newValue?.map((elem) => elem?.toData())?.toList()});
     return batch;
   }
 
   DocUpdate updateNotes(firestore.Firestore fs, String documentPath, String newValue, [DocUpdate batch]) {
     notes = newValue;
-    batch ??= FsDocUpdate(fs.batch());
+    batch ??= FirestoreDocUpdate(fs.batch());
     batch.update(fs.doc(documentPath), data: {'notes': newValue});
     return batch;
   }
 
   DocUpdate updateUnread(firestore.Firestore fs, String documentPath, bool newValue, [DocUpdate batch]) {
     unread = newValue;
-    batch ??= FsDocUpdate(fs.batch());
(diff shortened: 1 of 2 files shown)


webapp/lib/model.g.dart
```diff
@@ -417,12 +417,43 @@ void listenForUpdates<T>(
...
   final Map<String, dynamic> data;
 
   DocSnapshot(this.id, this.data);
 }
 
+/// A batch update, used to perform multiple writes as a single atomic unit.
+/// None of the writes are committed (or visible locally) until
+/// [DocUpdate.commit()] is called.
+abstract class DocUpdate {
+  /// Commits all of the writes in this write batch as a single atomic unit.
```
lukechurch: They're not atomic - that would only be the case if we were using transactions - could you adjust the comments here  to 'batch'?

webapp/lib/model.g.dart
```diff
@@ -417,12 +417,43 @@ void listenForUpdates<T>(
...
+  /// Note that it won't resolve while you're offline.
+  Future<Null> commit();
+
+  /// Updates fields in the document referred to by this [DocumentReference].
+  /// The update will fail if applied to a document that does not exist.
+  void update(firestore.DocumentReference doc, {Map<String, dynamic> data});
+}
+
+/// A batch update for documents in firestore.
+class FsDocUpdate implements DocUpdate {
```
lukechurch: I think we need to expand 'Fs' - It mean 'filesystem' to me first off. Could we Go with the full verbose thing of FireStore?

webapp/lib/platform.dart
```diff
@@ -138,13 +138,13 @@ Future updateUnread(List<Conversation> conversations, bool newValue) async {
       : "${conversations.length} conversations"
   }");
   if (conversations.isEmpty) return null;
-  var batch = _firestoreInstance.batch();
+  var batch = FsDocUpdate(_firestoreInstance.batch());
   int batchSize = 0;
```
lukechurch: Where does batch size get updated here? It seems to always be zero?

webapp/lib/model.g.dart
```diff
@@ -417,12 +417,43 @@ void listenForUpdates<T>(
...
   final String id;
   final Map<String, dynamic> data;
 
   DocSnapshot(this.id, this.data);
 }
 
+/// A batch update, used to perform multiple writes as a single atomic unit.
+/// None of the writes are committed (or visible locally) until
+/// [DocUpdate.commit()] is called.
+abstract class DocUpdate {
```
lukechurch: I wonder if this should be called `DocUpdateBatch`? It wasn't obvious to me from the name that it was really doing batching

webapp/lib/model.g.dart
danrubel: @lukechurch  According to the [WriteBatch online docs](https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/WriteBatch), it is atomic.

Please let me know how you would like to proceed and I can update the comment in a future PR.

webapp/lib/model.g.dart
danrubel: Good point. Done.

webapp/lib/platform.dart
danrubel: Thanks! Good catch! Fixed.

webapp/lib/model.g.dart
danrubel: Good point. `DocUpdateBatch` is more indicative of its function. Renamed.

PR #190, opened 2020-01-22
Add sample workflow file for automating issues and PRs

marianamarasoiu:
A proposal GitHub workflow for automating the generation of project cards when assigning issues, opening PRs and requesting PR reviews.

I'm not super happy with the solution, but it also seems that this is the best that is available at the moment. Perhaps in the future GitHub will also add automation to adding cards to projects, not just inside projects. In the meantime, please have a look at the code, and here are some of my concerns:

Since there doesn't seem to be a syntax for loops in GitHub workflows, we'll need to list each project/user separately - I was thinking each of us could have our own file.

Another potential downside is that these workflows run per repo, so we would need to import them into each (old and new) repo. I'm not entirely sure whether it's worth littering the codebases with this, and also whether the 'almost' automation that results from this is likely to generate frustration when issues/PRs get ignored because they're not being automatically added to the project.

Since the repos are under larksystems, the workflow needs permissions to run through a personal access token that is stored as a secret on the repo. It seems that we only need one per repo, and so the process of adding one will only need to happen the first time we add a workflow to a repo. (info about this [here](https://github.com/marketplace/actions/github-project-automation)).

Thanks and let me know what you think!

PS. To test this, I've created a separate repo with no code, just the workflows: https://github.com/larksystems/github-actions-test

Feel free to play with it as well :)

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,7 @@
+# GitHub Project Automation
+
+If you'd like to automate card creation from GitHub issues and pull requests for this repository using [GitHub Actions](https://help.github.com/en/actions), you can create your own workflow `.yml` file with your name in this folder.
+
+The easiest way to get started is to copy one of the existing files in this folder and replace your name/username throughout the file. Depending on your specific project layout, you may also want to check that the column names are correct.
+
+This workflow uses the [GitHub Project Automation+](https://github.com/marketplace/actions/github-project-automation) Action in order to create the cards in the specified project and column.
--- a/.github/workflows/mariana.yml
+++ b/.github/workflows/mariana.yml
@@ -1,19 +1,40 @@
-name: Move assigned issues and review requests
+name: Add assigned issues and review requests to Mariana's project
 
 on:
   issues:
     types: [assigned]
   pull_request:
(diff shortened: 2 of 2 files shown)


marianamarasoiu:
Thanks @danrubel !

I think forking `alex-page/github-project-automation-plus@v0.1.2` is likely to be what we're heading for, as we might want to support other project automations. For example, at the moment, removing a reviewer doesn't remove the card from their board.

Even automation inside the board could be helped - if a PR has two reviewers, after one of them approves the PR, the cards move into _Reviewer approved_ (if using the boards are using the default PR reviewer automation) for everyone, including the reviewer who hasn't approved the PR yet. If the PR author is expecting both reviewers to review the PR, then the automation is actually getting in the way. (We might want to first discuss what adding two reviewers means in Lark repos, though I would rather have the PR author decide to merge with 1/2 reviews rather than automation deciding for us)

I like the sound of having a `larksystems/github-dashboard-automation` action repo with a single workflow file per repo. I'll have a look at what it would take to set that up.

Thanks again Dan, I'm going to merge this now, even if it might get replaced soon, just so we can get a bit more testing out of it :)

PR #191, opened 2020-01-23
move firebase specific functionality out of model

danrubel:
This PR finishes moving Firebase specific functionality out of both `model.dart` and `model.g.dart` and into a new `model_firebase.dart` file. This is a step closer to driving nook tests with json rather than firebase.

Fix https://github.com/larksystems/KK-Project-2019-WorldBank-PLR/issues/31


lukechurch:
@marianamarasoiu - Could I ask you to take a look at this?

PR #192, opened 2020-01-28
new "after date" conversation filter

danrubel:
This PR adds a new "after date" filter to show only conversations whose last message has a timestamp more recent than a specified date/time.

after date filter 1

When selected, a new text area appears in which a date / time can be entered. The format is YYYY MM DD HH. The year is required and must be a contiguous group of 4 digits at the start, but the month, day, and hour are optional and all non-numeric characters entered in this text field will be ignored when parsing the date/time.

after date filter 2

When the "Apply" button is clicked or the "Enter" key is pressed, the text is parsed into a date/time and applied as a filter

after date filter 3

### DIFF
--- a/webapp/lib/view.dart
+++ b/webapp/lib/view.dart
@@ -185,6 +185,13 @@ class ConversationPanelView {
       _messages.firstChild.remove();
     }
   }
+
+  void showAfterDateFilterPrompt(DateTime dateTime) {
+    log.warning("showAfterDateFilterPrompt");
+    // TODO prompt for after date filter
+    dateTime ??= DateTime.now();
+    command(UIAction.updateAfterDateFilter, new AfterDateFilterData(AFTER_DATE_TAG_ID, dateTime));
+  }
 }
 
 class MessageView {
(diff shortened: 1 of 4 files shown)


lukechurch:
Mariana, any chance you could take a look at this this morning? Its not necessarily intended as a long term solution, but the lack of date filtering is causing an operations problem on one of the production uses of katikati at the moment.

webapp/lib/view.dart
```diff
@@ -185,6 +189,104 @@ class ConversationPanelView {
...
+          e.stopImmediatePropagation();
+          applyFilter();
+        }
+      });
+
+    panel = DivElement()
+      ..classes.add('after-date-prompt')
+      ..append(SpanElement()
+        ..classes.add('after-date-prompt__prompt-text')
+        ..text = 'Enter date:')
```
marianamarasoiu: Could we a bit of additional text here for the expected date format?

danrubel: Done.

PR #193, opened 2020-01-29
model cleanup

danrubel:
Two cleanups and one fix:
* move Conversation.collectionName into model
* remove datetime default
* fix batch update unread flag

### DIFF
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -88,7 +88,7 @@ class Message {
     if (data == null) return null;
     return (modelObj ?? Message())
       ..direction = MessageDirection.fromString(data['direction'] as String) ?? MessageDirection.Out
-      ..datetime = DateTime_fromData(data['datetime']) ?? DateTime.now()
+      ..datetime = DateTime_fromData(data['datetime'])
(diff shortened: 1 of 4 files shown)


lukechurch:
Handing this one to @marianamarasoiu for a first look

webapp/lib/model.g.dart
```diff
@@ -85,7 +88,7 @@ class Message {
     if (data == null) return null;
     return (modelObj ?? Message())
       ..direction = MessageDirection.fromString(data['direction'] as String) ?? MessageDirection.Out
-      ..datetime = DateTime_fromData(data['datetime']) ?? DateTime.now()
+      ..datetime = DateTime_fromData(data['datetime'])
```
marianamarasoiu: Is this because `datetime` is now expected to be mandatory as a field?

danrubel: @marianamarasoiu @lukechurch Correct. All fields will eventually be mandatory, but it will take some time to get there. When new field are introduced (e.g. MessageStatus), they will have a default value until they can be fully plumbed through the back end and all existing data migrated to have a value. Once that happens, the default value can be removed.

Issue #189, opened 2020-01-21
Tag filtering needs more expressivity


marianamarasoiu:
Currently the only operation between tags is AND. We also need OR and NOT.

PR #194, opened 2020-02-03
Add initial separation between include and exclude filters

marianamarasoiu:
Re #189

This is a first step for having two filter lists - one for tags to search for, another for tags to exclude, plus an AND/OR operator on each list.

Currently the exclude list is empty and not functional, that's coming up next.

cc @lukechurch for feedback on the UI approach

Screenshots:
Screenshot 2020-02-03 at 12 14 58
Screenshot 2020-02-03 at 12 15 25


marianamarasoiu:
Thanks @danrubel !
I had a quick offline chat with @lukechurch and this seems to be heading in the right direction.

Since this is still a WIP and Nook needs to work sensibly at `HEAD` on `master`, I've created a `dev-filtering` branch which I'll be using as the base for this and future PRs until the filtering feature is ready to be merged into master.

PR #195, opened 2020-02-06
Update unread flag via pub/sub

danrubel:
This changes the underlying model so that changes to the `Conversation` `unread` flag in firebase are made via pub/sub rather than directly updating firebase.

* `setUnread` - update the `unread` flag for a single conversation
* `setAllUnread` - update the `unread` flag for multiple conversations

A new `DocPubSubUpdate` class (interface) has been added so that the model does not have to know the specifics behind pub/sub implementation.

For background see [reroute nook actions through pub/sub](https://github.com/larksystems/KK-Project-2020-IOM/issues/18)

### DIFF
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -10,6 +10,7 @@ Logger log = Logger('model.g.dart');
 class Conversation {
   static const collectionName = 'nook_conversations';
 
+  String docId;
   Map<String, String> demographicsInfo;
   List<String> tagIds;
   List<Message> messages;
@@ -17,7 +18,7 @@ class Conversation {
   bool unread;
 
   static Conversation fromSnapshot(DocSnapshot doc, [Conversation modelObj]) =>
-      fromData(doc.data, modelObj);
+      fromData(doc.data, modelObj)..docId = doc.id;
 
   static Conversation fromData(data, [Conversation modelObj]) {
     if (data == null) return null;
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -71,6 +71,16 @@ class Conversation {
     batch.update(documentPath, data: {'unread': newValue});
     return batch;
   }
+
+  void setUnread(PubSubClient pubSubClient, bool newValue) {
+    unread = newValue;
+    pubSubClient.publish(platform_constants.smsTopic, {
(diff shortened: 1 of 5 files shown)


webapp/lib/model.g.dart
```diff
@@ -64,11 +65,20 @@ class Conversation {
     return batch;
   }
 
-  DocBatchUpdate updateUnread(DocStorage docStorage, String documentPath, bool newValue, [DocBatchUpdate batch]) {
-    unread = newValue;
-    batch ??= docStorage.batch();
-    batch.update(documentPath, data: {'unread': newValue});
-    return batch;
+  Future<bool> setUnread(DocPubSubUpdate pubSubClient, bool newValue) {
```
lukechurch: for future PR: doc comment about what the boolean is?

webapp/lib/model.g.dart
```diff
@@ -447,3 +457,10 @@ abstract class DocBatchUpdate {
 
 // ======================================================================
 // Core pub/sub utilities
+
+/// A pub/sub based mechanism for updating documents
+abstract class DocPubSubUpdate {
+  /// Publish the given document changes,
+  /// where [changes] is a mapping of field name to new value
+  Future<bool> publishDocChange(String collectionName, List<String> docIds, Map<String, dynamic> changes);
```
lukechurch: Can you delete fields with this? I think we'll need to be able to store nulls so we might need to seperate into 'changes' and 'deletes'?

danrubel: @lukechurch Good point. I think we'll need `setField` and `clearField` along with `addToField` and `removeFromField` for fields that are lists (e.g. tags). Will address in a subsequet PR.

webapp/lib/model.g.dart
danrubel: @lukechurch Happy to add a comment but I'm not understanding exactly what you'd like added. What comment would you like added in this specific case? Maybe then I'll understand what you are looking for in general.

danrubel: @lukechurch After some time and rereading this comment, I understand your concern. You are asking about the `bool` return type not the `bool` argument type. After pondering this, I've changed the return type to `Future<void>` in https://github.com/larksystems/nook/pull/217

PR #197, opened 2020-02-08
redirect Conversation.notes to be updated via pub/sub

danrubel:
See https://github.com/larksystems/KK-Project-2020-IOM/issues/18

### DIFF
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -58,11 +58,20 @@ class Conversation {
     return batch;
   }
 
-  DocBatchUpdate updateNotes(DocStorage docStorage, String documentPath, String newValue, [DocBatchUpdate batch]) {
-    notes = newValue;
-    batch ??= docStorage.batch();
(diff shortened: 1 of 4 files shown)


webapp/lib/model.g.dart
```diff
@@ -58,11 +58,20 @@ class Conversation {
     return batch;
   }
 
-  DocBatchUpdate updateNotes(DocStorage docStorage, String documentPath, String newValue, [DocBatchUpdate batch]) {
-    notes = newValue;
-    batch ??= docStorage.batch();
-    batch.update(documentPath, data: {'notes': newValue});
-    return batch;
+  Future<bool> setNotes(DocPubSubUpdate pubSubClient, String newValue) {
```
lukechurch: Consider calling this setNote or setConversationNote - the plurality here was making it a bit harder tor me to understand the difference between it and the set all code below

danrubel: @lukechurch The field is called `notes`, thus `set<name-of-field>`... should we rename the field? Or just this method?

PR #200, opened 2020-02-11
prevent text edit keystrokes from triggering actions

danrubel:
This updates areas that are editable so that text editing keystrokes in that area do not also trigger messages to be sent or tags to be applied.


danrubel:
@marianamarasoiu Good point. I thought about wrapping all the editable elements into new classes, but I was unsure and thus chose this path instead. Happy to refactor further into editable elements if the need arises.

PR #203, opened 2020-02-12
redirect reply.translation to be updated via pub/sub 

danrubel:
This builds on the prior [refactor save text action PR](https://github.com/larksystems/nook/pull/202) so that reply.translation changes are consolidated over a rolling 3 second period and updated via pub/sub rather than directly modifying firebase.

In addition, 2 unused `Message` methods were removed.

cc @lukechurch

### DIFF
--- a/webapp/lib/model.yaml
+++ b/webapp/lib/model.yaml
@@ -7,6 +7,7 @@ Conversation:
   unread: 'publishable bool, true'
 
 Message:
+  firebaseDocId: 'none'
   direction: 'MessageDirection, MessageDirection.Out'
   datetime: 'datetime'
   status: MessageStatus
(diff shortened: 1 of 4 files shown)


webapp/lib/model.yaml
```diff
@@ -7,6 +7,7 @@ Conversation:
   unread: 'publishable bool, true'
 
 Message:
+  firebaseDocId: 'none'
```
marianamarasoiu: What does this do?

danrubel: This tells codegen that `Message` is not a document but rather an element in a document and thus the `fromSnapshot` and `listen` methods need not be generated. I've updated the [model codegen doc](https://docs.google.com/document/d/1Jnr6TRjZFIlyw_pVXIwB58aGmUxkcWWYasOVt_jL42E/edit) with this information, but sometime soon I need to move this into a markup file in the `Infrastructure` repo

marianamarasoiu: Ooo a design doc, sweet :)
Great, thanks for this, that makes sense!

PR #207, opened 2020-02-16
Bump dart SDK version so we can use extension methods

marianamarasoiu:
I'm using extension methods in an upcoming PR - if you don't think it's a good idea to go to Dart 2.7, please let me know and I'm happy to go back.

Thanks!


marianamarasoiu:
thanks @danrubel! Sorry for taking a while merging this, I tried to do it yesterday but Github kept timing out 🤷‍♀

PR #208, opened 2020-02-16
Separate between include and exclude filters

marianamarasoiu:
Continuing work on more expressive filtering #189 

At the moment, only the tags are considered for the OR/AND operation, the after date is always an AND, even though it's not differentiated in the UI as such. Based on the use cases discussed so far, I think it makes sense to show that its effect is different - @lukechurch can you confirm that this makes sense?

Thanks (and sorry for the slightly extensive change)!

Mini demo:
there are 2 conversations, tagged as follows:

conv1 - tag1
conv2 - tag2

show *any* tag1 tag2:
Screenshot 2020-02-16 at 21 48 25

show *all* tag1 tag2:
Screenshot 2020-02-16 at 21 48 43

hide *all* tag1:
Screenshot 2020-02-16 at 21 49 05

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -157,7 +157,8 @@ class ConversationFilter {
   FilterOperation includeLogic = FilterOperation.all;
   List<model.Tag> excludeTags = [];
   FilterOperation excludeLogic = FilterOperation.all;
-  DateTime afterDateFilter;
+  DateTime includeAfterDateFilter;
+  DateTime excludeAfterDateFilter;
 
   static final defaultIncludeLogic = FilterOperation.all;
   static final defaultExcludeLogic = FilterOperation.all;
@@ -177,7 +178,6 @@ List<model.SuggestedReply> suggestedReplies;
 List<model.Tag> conversationTags;
 List<model.Tag> messageTags;
 ConversationFilter conversationFilter;
-DateTime afterDateFilter;
 model.Conversation activeConversation;
 List<model.Conversation> selectedConversations;
 model.Message selectedMessage;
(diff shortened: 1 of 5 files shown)


webapp/lib/controller.dart
```diff
@@ -690,16 +778,45 @@ void setMessageTag(model.Tag tag, model.Message message, model.Conversation conv
   }
 }
 
-Set<model.Conversation> filterConversationsByTags(Set<model.Conversation> conversations, List<model.Tag> filterTags, DateTime afterDateFilter) {
-  if (filterTags.isEmpty && afterDateFilter == null) return conversations;
+Set<model.Conversation> filterConversationsByTags(Set<model.Conversation> conversations, ConversationFilter conversationFilter) {
+  if (conversationFilter.includeTags.isEmpty &&
+      conversationFilter.excludeTags.isEmpty &&
+      conversationFilter.includeAfterDateFilter == null &&
+      conversationFilter.excludeAfterDateFilter == null) {
```
danrubel: Perhaps `if (conversationFilter.isEmpty)` or something like that? 
I was looking for ways to make code easier to read by adding utility methods to the `ConversationFIlter` class.

webapp/lib/controller.dart
```diff
@@ -690,16 +778,45 @@ void setMessageTag(model.Tag tag, model.Message message, model.Conversation conv
...
+    if (conversationFilter.excludeTags.isNotEmpty) {
+      switch (conversationFilter.excludeLogic) {
+        case FilterOperation.all:
+          if (conversation.tagIds.toSet().containsAll(excludeFilterTagIds)) return;
+          break;
+        case FilterOperation.any:
+          if (conversation.tagIds.toSet().intersection(excludeFilterTagIds).isNotEmpty) return;
+          break;
+      }
+    }
```
danrubel: Would it make sense moving this logic into a `ConversationFilter.filterConversation(conversation)` method?

webapp/lib/controller.dart
marianamarasoiu: Great idea, thanks!

webapp/lib/controller.dart
marianamarasoiu: I agree, I think that makes sense. Done, thanks!

marianamarasoiu:
Thanks @danrubel , merging now!

PR #211, opened 2020-02-22
update Conversation.tagIds to use pub/sub add/remove

danrubel:
This updates `Conversation.tagIds` to use the new pub/sub based add to and remove from list operations. In addition `setAll<fieldName>` is renamed to `set<fieldName>_forAll`

See [reroute nook actions through pub/sub](https://github.com/larksystems/KK-Project-2020-IOM/issues/18)

### DIFF
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -44,20 +44,20 @@ class Conversation {
     };
   }
 
-  Future<bool> addToTagIds(DocPubSubUpdate pubSubClient, String newValue) {
-    return addToTagIds_forAll(pubSubClient, [this], newValue);
+  Future<bool> addToTagId(DocPubSubUpdate pubSubClient, String valueToAdd) {
+    return addToTagId_forAll(pubSubClient, [this], valueToAdd);
   }
 
-  static Future<bool> addToTagIds_forAll(DocPubSubUpdate pubSubClient, List<Conversation> docs, String newValue) async {
-    final docIds = <String>[];
+  static Future<bool> addToTagId_forAll(DocPubSubUpdate pubSubClient, List<Conversation> docs, String valueToAdd) async {
+    final docIdsToPublish = <String>[];
     for (var doc in docs) {
-      if (!doc.tagIds.contains(newValue)) {
-        doc.tagIds.add(newValue);
-        docIds.add(doc.docId);
+      if (!doc.tagIds.contains(valueToAdd)) {
+        doc.tagIds.add(valueToAdd);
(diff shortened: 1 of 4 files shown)


webapp/lib/model.g.dart
```diff
@@ -60,6 +76,22 @@ class Conversation {
     return pubSubClient.publishDocChange(collectionName, docIds, {"tags": newValue?.toList()});
   }
 
+  Future<bool> removeFromTagIds(DocPubSubUpdate pubSubClient, String newValue) {
```
lukechurch: I'm not sure about 'newValue' as the argument name - is it the name of the argument to remove?

webapp/lib/model.g.dart
```diff
@@ -44,11 +44,27 @@ class Conversation {
     };
   }
 
+  Future<bool> addToTagIds(DocPubSubUpdate pubSubClient, String newValue) {
+    return addToTagIds_forAll(pubSubClient, [this], newValue);
+  }
+
+  static Future<bool> addToTagIds_forAll(DocPubSubUpdate pubSubClient, List<Conversation> docs, String newValue) async {
+    final docIds = <String>[];
```
marianamarasoiu: nit: rename to `docIdsToPublish` ?

webapp/lib/model.g.dart
```diff
@@ -44,11 +44,27 @@ class Conversation {
     };
   }
 
+  Future<bool> addToTagIds(DocPubSubUpdate pubSubClient, String newValue) {
```
marianamarasoiu: nit: I think `addTagId` sounds better? unless `newValue` can be something else other than a tag ID?

webapp/lib/model.g.dart
marianamarasoiu: same as above - consider renaming to `removeTagId`.

danrubel: Good catch. Renamed to `valueToRemove` here and elsewhere.

webapp/lib/model.g.dart
danrubel: I like that better. Fixed.

webapp/lib/model.g.dart
danrubel: Good point. Fixed.

webapp/lib/model.g.dart
danrubel: Good point. Fixed.

danrubel:
@marianamarasoiu Thanks for the suggestions!

Issue #162, opened 2019-12-13
Delayed message sending (Measure twice, cut once)


lukechurch:
It's a bit scary to be able to send so many messages so quickly - and with no undo. Esp. from the keyboard shortcuts.

Consider building an 'outbox' facility that buffers the outgoing messages so they can be reviewed before being sent

marianamarasoiu:
Instant message sending is also an issue when replying from a different timezone - sending messages in the middle of the night is not appropriate.

The 'outbox' facility should also include the ability to schedule messages to be sent.

PR #215, opened 2020-02-27
Add support for two views in preparation of adding an outbox feature

marianamarasoiu:
Re #162 

This creates a top menu for switching between the conversation view (the UI we already have) and the outbox, an upcoming new view for managing outgoing messages (delayed sending to allow double checking, scheduled sending etc.)

As with the filtering feature, because it's a substantial change and intermediate versions might not work, I'm working on a separate branch off master (`dev-outbox`) which when ready, I'll merge into master.

cc @lukechurch for following progress and feedback on UI design

Demo: 
Screenshot 2020-02-27 at 16 22 02
Screenshot 2020-02-27 at 16 22 42

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -194,7 +194,6 @@ void initUI() {
       conversationTags.removeWhere((tag) => updatedIds.contains(tag.tagId));
       conversationTags.addAll(tags);
       _populateFilterTagsMenu(conversationTags);
-      print(actionObjectState);
       if (actionObjectState == UIActionObject.conversation) {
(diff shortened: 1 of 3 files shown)


webapp/lib/controller.dart
```diff
@@ -438,15 +470,15 @@ void command(UIAction action, Data data) {
     case UIAction.userSignedOut:
```
danrubel: Should "universal" actions such as sign out be moved into the `command` function above?

marianamarasoiu: That would make sense, yes. I'll address this in a separate PR, thanks!

marianamarasoiu:
thanks @danrubel , merging now..

PR #217, opened 2020-02-28
pub/sub exception handling

danrubel:
This updates the pub/sub return value and generated exceptions in the model. Specifically:
* change pub/sub methods to return `Future<void>` rather than `Future<bool>`
* change pub/sub mechanism to throw PubSubException if return status code != OK
* update controller to catch pub/sub exceptions and display a snackbar message to alert the user rather than just silently failing
image

### DIFF
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -44,64 +44,64 @@ class Conversation {
     };
   }
 
-  /// Add [valueToAdd] to tagIds in the receiver.
+  /// Add [tagId] to tagIds in this Conversation.
   /// Callers should catch and handle IOException.
-  Future<void> addToTagId(DocPubSubUpdate pubSubClient, String valueToAdd) {
-    return addToTagId_forAll(pubSubClient, [this], valueToAdd);
+  Future<void> addTagId(DocPubSubUpdate pubSubClient, String tagId) {
+    return addTagIdToAll(pubSubClient, [this], tagId);
   }
 
-  /// Add [valueToAdd] to tagIds for each Conversation.
+  /// Add [tagId] to tagIds in each Conversation.
   /// Callers should catch and handle IOException.
-  static Future<void> addToTagId_forAll(DocPubSubUpdate pubSubClient, List<Conversation> docs, String valueToAdd) async {
+  static Future<void> addTagIdToAll(DocPubSubUpdate pubSubClient, List<Conversation> docs, String tagId) async {
     final docIdsToPublish = <String>[];
     for (var doc in docs) {
-      if (!doc.tagIds.contains(valueToAdd)) {
-        doc.tagIds.add(valueToAdd);
+      if (!doc.tagIds.contains(tagId)) {
+        doc.tagIds.add(tagId);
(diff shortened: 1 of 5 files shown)


webapp/lib/model.g.dart
```diff
@@ -44,51 +44,63 @@ class Conversation {
...
+  /// Remove [valueToRemove] from tagIds in the receiver.
+  /// Callers should catch and handle IOException.
+  Future<void> removeFromTagId(DocPubSubUpdate pubSubClient, String valueToRemove) {
     return removeFromTagId_forAll(pubSubClient, [this], valueToRemove);
   }
 
-  static Future<bool> removeFromTagId_forAll(DocPubSubUpdate pubSubClient, List<Conversation> docs, String valueToRemove) async {
+  /// Remove [valueToRemove] from tagIds in each Conversation.
+  /// Callers should catch and handle IOException.
+  static Future<void> removeFromTagId_forAll(DocPubSubUpdate pubSubClient, List<Conversation> docs, String valueToRemove) async {
```
marianamarasoiu: I think this should be named either `removeTagId` or `removeFromTagIds`? I have a slight preference for the former option, but your call.

webapp/lib/model.g.dart
```diff
@@ -99,35 +111,43 @@ class Conversation {
     return batch;
   }
 
-  Future<bool> setNotes(DocPubSubUpdate pubSubClient, String newValue) {
+  /// Set notes for the receiver.
```
marianamarasoiu: s/receiver/conversation ?

marianamarasoiu: Actually I see now that all the other non-static methods use receiver. Can we replace it with the respective object name/type? thanks!

webapp/lib/pubsub.dart
```diff
@@ -18,7 +20,9 @@ class PubSubClient extends DocPubSubUpdate {
 
   PubSubClient(this.publishUrl, this.user);
 
-  Future<bool> publish(String topic, Map payload) async {
+  /// Publish the specified exception.
+  /// Callers should catch and handle IOException.
+  Future<void> publish(String topic, Map payload) async {
```
marianamarasoiu: does this still need to be async?

webapp/lib/model.g.dart
danrubel: Good catch. I need to distinguish between removing a tagId from the list of tagIds in a single conversation and doing the same in multiple conversations, and I've been struggling what to name the latter. I've changed them to
* `addTagId` - add a single tag id to the tagIds in this Conversation
* `addTagIdToAll` - add a single tag id to the tagIds in each specified Conversation
* `setTagIds` - set the tagIds in this Conversation
* `setTagIdsForAll - set the tagIds for each specified Conversation
* `removeTagId` - remove a single tag id from the tagIds in this Conversation
* `removeTagIdFromAll` - remove a single tag id from the tagIds in each specified Conversation
Happy to discuss and rename again :-)

webapp/lib/model.g.dart
danrubel: Good idea. Fixed.

PR #221, opened 2020-02-28
Update message tags and translation via pub/sub

danrubel:
This updates the controller, platform, and model to use pub/sub when
* adding a message tag
* removing a message tag
* updating a message translation

In addition, updating a message translation now uses `SaveTextAction` similar to updating a conversation note.

Fix https://github.com/larksystems/KK-Project-2020-IOM/issues/18

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -425,15 +425,23 @@ void command(UIAction action, Data data) {
         );
       } else if (data is TranslationData) {
         TranslationData messageTranslation = data;
-        var message = activeConversation.messages[messageTranslation.messageIndex];
-        platform.setMessageTranslation(activeConversation, message, messageTranslation.translationText);
+        var conversation = activeConversation;
+        var message = conversation.messages[messageTranslation.messageIndex];
+        SaveTextAction.textChange(
+          "${conversation.docId}.message-${messageTranslation.messageIndex}.translation",
+          messageTranslation.translationText,
+          (newText) {
+            return platform.setMessageTranslation(conversation, message, newText);
+          },
+        );
       }
       break;
     case UIAction.updateNote:
+      var conversation = activeConversation;
       SaveTextAction.textChange(
-        "${activeConversation.docId}.notes",
+        "${conversation.docId}.notes",
         (data as NoteData).noteText,
-        (newText) => platform.updateNotes(activeConversation, newText),
+        (newText) => platform.updateNotes(conversation, newText),
       );
       break;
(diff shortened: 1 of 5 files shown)


webapp/lib/controller.dart
```diff
@@ -426,16 +425,26 @@ void command(UIAction action, Data data) {
...
       }
       break;
-    case UIAction.updateNote:
-      SaveTextAction.textChange(
-        "${activeConversation.docId}.notes",
-        (data as NoteData).noteText,
-        (newText) => platform.updateNotes(activeConversation, newText),
-      );
+    case UIAction.updateNote: {
+        // inside block to ensure conversation is a local/captured var
```
marianamarasoiu: why is this an issue? and if it is, would you need to have an inside block for the above code as well?

webapp/lib/model.dart
```diff
@@ -44,7 +44,57 @@ extension ConversationUtil on g.Conversation {
   }
 }
 
-UnmodifiableListView<g.Tag> tagIdsToTags(Iterable<String> tagIds, Iterable<g.Tag> allTags) {
+extension MessageUtil on g.Message {
+  /// Add [tagId] to tagIds in this Message.
+  /// Callers should catch and handle IOException.
+  Future<void> addTagId(g.DocPubSubUpdate pubSubClient,
+      g.Conversation conversation, String tagId) async {
```
marianamarasoiu: This is more of a style question, but I find the split on two lines here a bit hard to read (same for the other methods below). Is this because of a linter being a bit eager on line length? Would it be possible to disable that check? thanks!

webapp/lib/controller.dart
danrubel: `activeConversation` is a global which can change before the `(newText) => ...` closure executes. Thus I assign `activeConversation` to a local var `conversation` to capture the value.

On hindsight, the `{ }` are not necessary and can be removed. I'll tackle that in a subsequent PR.

webapp/lib/model.dart
danrubel: Good point. I'll work on readability for these three methods in a subsequent CL.

webapp/lib/controller.dart
danrubel: Fixed in https://github.com/larksystems/nook/pull/222

webapp/lib/model.dart
danrubel: Fixed in https://github.com/larksystems/nook/pull/222

PR #223, opened 2020-03-10
Remove reference to dart:io libary from web platform code

marianamarasoiu:
I think the `dart:io` library got introduced accidentally In #217 and I didn't catch it in review either, resulting in the app not compiling any more.

Is there a particular reason why `PubSubException` should inherit from `IOException`, or can it inherit directly from `Exception`? If it's not too important, here's a simple PR to remove the `dart:io` imports and remove `IOException`, but if you would like to do this a different way, happy to discuss.

Thanks!

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -2,7 +2,6 @@ library controller;
 
 import 'dart:async';
 import 'dart:collection';
-import 'dart:io';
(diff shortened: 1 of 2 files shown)


marianamarasoiu:
No worries, and thanks @danrubel , merging now!

PR #228, opened 2020-03-13
Add a toggle in the tags panel for hiding age tags

marianamarasoiu:
Demo screenshot:

Screenshot 2020-03-13 at 15 21 02

@lukechurch if you'd like to hide other types of tags, plmk

Thanks!

### DIFF
--- a/webapp/lib/view.dart
+++ b/webapp/lib/view.dart
@@ -828,6 +828,7 @@ class TagPanelView {
   DivElement _tags;
   DivElement _tagList;
   DivElement _statusPanel;
+  InputElement _hideTagsCheckbox;
   Text _statusText;
 
   AddActionView _addTag;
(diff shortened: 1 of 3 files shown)


webapp/lib/view.dart
```diff
@@ -870,6 +885,25 @@ class TagPanelView {
...
+    for(DivElement tag in _tagList.children) {
+      if (!showAll && isAgeTag(tag)) {
+        tag.classes.toggle('action--hide', true);
+        continue;
+      }
+      tag.classes.toggle('action--hide', false);
+    }
+  }
+
+  bool isAgeTag(DivElement tag) {
```
lukechurch: This feels like a bit of a workaround to a proper tagging management system - all fine, but could you add a comment describing it as such?

marianamarasoiu: Thanks @lukechurch , added a comment, please have another look

PR #241, opened 2020-03-15
Add optional sequence number to suggested replies

marianamarasoiu:


marianamarasoiu:
@danrubel - feel free to review after PR merge in case we need to merge this before you get the chance to have a look at it. thanks!

webapp/lib/controller_view_helper.dart
```diff
@@ -59,7 +59,11 @@ void _populateConversationPanelView(model.Conversation conversation) {
 }
 
 void _populateReplyPanelView(List<model.SuggestedReply> replies) {
-  replies.sort((r1, r2) => r1.shortcut.compareTo(r2.shortcut));
+  replies.sort((r1, r2) {
+    var seqNo1 = r1.seqNumber == null ? double.nan : r1.seqNumber;
+    var seqNo2 = r2.seqNumber == null ? double.nan : r2.seqNumber;
+    return seqNo1.compareTo(seqNo2);
```
danrubel: @marianamarasoiu If there are no sequence numbers (e.g. seqNo1 and seqNo2 are double.nan) then should we fall back on comparing shortcuts? Or will all suggested replies have sequence numbers?

danrubel:
@marianamarasoiu LGTM with one question

webapp/lib/controller_view_helper.dart
marianamarasoiu: Yup, we can fall back on comparing shortcuts . For the current project, I think all suggested replies will have sequence numbers and no shortcuts, so it wouldn't have much of an effect, but I think it's a good idea to have that in anyway. I'll prepare a PR to do that. thanks!

PR #243, opened 2020-03-15
Group suggested replies via a category field

marianamarasoiu:
A demo screenshot, with some test categories:

Screenshot 2020-03-15 at 14 06 29

@danrubel feel free to review this even if it's been merged, as we may need to merge it before you get the chance to have a look at it.

Thanks!

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -48,6 +48,7 @@ enum UIAction {
   enableMultiSelectMode,
   disableMultiSelectMode,
   updateSystemMessages,
+  updateSuggestedRepliesCategory
 }
 
 class Data {}
@@ -143,6 +144,11 @@ class AddTagData extends Data {
   AddTagData(this.tagText);
 }
 
+class UpdateSuggestedRepliesCategoryData extends Data {
+  String category;
+  UpdateSuggestedRepliesCategoryData(this.category);
+}
+
 class SystemMessagesData extends Data {
   List<model.SystemMessage> messages;
   SystemMessagesData(this.messages);
@@ -155,6 +161,8 @@ UIActionObject actionObjectState = UIActionObject.conversation;
 Set<model.Conversation> conversations;
 Set<model.Conversation> filteredConversations;
 List<model.SuggestedReply> suggestedReplies;
+Map<String, List<model.SuggestedReply>> suggestedRepliesByCategory;
+String selectedSuggestedRepliesCategory;
 List<model.Tag> conversationTags;
 List<model.Tag> messageTags;
 List<model.Tag> filterTags;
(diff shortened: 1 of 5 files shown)


webapp/lib/controller.dart
```diff
@@ -216,7 +225,36 @@ void initUI() {
         suggestedReplies.removeWhere((suggestedReply) => updatedIds.contains(suggestedReply.suggestedReplyId));
         suggestedReplies.addAll(updatedReplies);
 
-        _populateReplyPanelView(suggestedReplies);
+        // Update the replies by category map
+        suggestedRepliesByCategory = _groupRepliesIntoCategories(suggestedReplies);
+        // Replace null key (if exists) with empty string
+        if (suggestedRepliesByCategory.containsKey(null)) {
+          suggestedRepliesByCategory[''] = suggestedRepliesByCategory.remove(null);
```
danrubel: @marianamarasoiu  Will there ever be a null and empty string key together?
IOW will this line ever discard/overwrite a category?

webapp/lib/controller_view_helper.dart
```diff
@@ -131,3 +131,14 @@ view.TagStyle tagTypeToStyle(model.TagType tagType) {
...
   }
 }
+
+Map<String, List<model.SuggestedReply>> _groupRepliesIntoCategories(List<model.SuggestedReply> replies) {
+  Map<String, List<model.SuggestedReply>> result = {};
+  for (model.SuggestedReply reply in replies) {
+    if (!result.containsKey(reply.category)) {
+      result[reply.category] = [];
+    }
+    result[reply.category].add(reply);
```
danrubel: @marianamarasoiu If `reply.category` is either `null` or `""`, then should the reply be added to a `"default"` category instead? Then perhaps there would not need to be all the guard code in `controller.dart` lines 230 - 237

marianamarasoiu: Good point, the guard code can happen here, and we also avoid the problem of lists overwritting each other which you identified above. Thanks for the suggestion, I've now moved the null check here.

I considered naming it `"default"` but it feels a bit more awkward than the empty string (for example I would need to check again in the view to make sure the default is at the top in the list). I think empty category feels explicit enough? But let's see how it works in practice and if it's not clear, I'll use a `"default"`-type description.

webapp/lib/controller.dart
marianamarasoiu: Good catch, It is possible as category is an optional filed, so when both are present the null category group would overwrite the empty string group. I've now moved the check in the helper file where the grouping gets created, which also fixes this issue. Thank you!

marianamarasoiu:
thanks @danrubel and @lukechurch , merging this now

PR #244, opened 2020-03-15
Fallback to sorting by shortcut if sequence numbers are null

marianamarasoiu:
As discussed in #243 in comments.

Thanks!


marianamarasoiu:
Thanks @danrubel !

PR #246, opened 2020-03-16
Replace hacky view-only tag filtering

marianamarasoiu:
We've added a hide age-related tags checkbox in #228 , but it was a bit hacky and would only hide tags in the panel, not also in the tag filter menu.

This PR wires the checkbox to the controller, which can then propagate the hide/show decision in all the right places.

The way we check which tags to hide (ie. age by trying to parse the text) is still a bit hacky and we'll need to have a more sensible solution in the future.

I'm also not super fond of the names I was able to pick - if you have anything better than `_filterAgeTagsIfNeeded` plmk :)

Thanks!

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -48,7 +48,8 @@ enum UIAction {
   enableMultiSelectMode,
   disableMultiSelectMode,
   updateSystemMessages,
-  updateSuggestedRepliesCategory
+  updateSuggestedRepliesCategory,
+  hideAgeTags
 }
 
 class Data {}
@@ -154,6 +155,11 @@ class SystemMessagesData extends Data {
   SystemMessagesData(this.messages);
 }
 
+class ToggleData extends Data {
+  bool toggleValue;
+  ToggleData(this.toggleValue);
+}
+
 List<model.SystemMessage> systemMessages;
 
 UIActionObject actionObjectState = UIActionObject.conversation;
@@ -172,6 +178,7 @@ List<model.Conversation> selectedConversations;
 model.Message selectedMessage;
 model.User signedInUser;
 
+bool hideAgeTags;
 bool multiSelectMode;
 
 void init() async {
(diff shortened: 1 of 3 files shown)


webapp/lib/controller.dart
```diff
@@ -610,6 +619,27 @@ void command(UIAction action, Data data) {
       selectedSuggestedRepliesCategory = updateCategoryData.category;
       _populateReplyPanelView(suggestedRepliesByCategory[selectedSuggestedRepliesCategory]);
       break;
+    case UIAction.hideAgeTags:
+      ToggleData toggleData = data;
+      hideAgeTags = toggleData.toggleValue;
+      switch (actionObjectState) {
+        case UIActionObject.conversation:
```
danrubel: Should age tags be un/hidden in both conversations and message together? Just trying to understand when we want age tags to be hidden in conversations but visible in messages, or vise versa.

marianamarasoiu: Yes, we want tags hidden both in conversation and in messages. However, in the UI, only one set will be visible at a time, so depending on which one we're showing (ie. do we have an entire conversation selected or just a single message, which is given by `actionObjectState`), we need to update the panel listing tags on the RHS with either conversation tags or message tags.

The menu for filtering conversations is based on only conversation tags (though maybe in the future this will change). Hence the need to compute `filteredConversationTags` even when showing message tags in the panel.

I opted for a transient way to filter tags, ie. compute filtered messages as needed rather than keeping track of specifically which set is being displayed at any moment, as this seemed easier and wouldn't add any new state to keep track of.

Rereading the code I think it would make it clearer if I just computed both at the top followed by the switch to populate the panel, followed by populating the menu, so I've went ahead and tidied it up like that.

Does this make more sense now? If not, please let me know!

danrubel: @marianamarasoiu Ah! I didn't understand that only one set at a time would be visible in the UI. All good!

PR #248, opened 2020-03-16
Remove Unicef logo from front page

marianamarasoiu:


webapp/lib/view.dart
```diff
@@ -1266,11 +1266,6 @@ class AuthMainView {
       ..classes.add('partner-logo--avf');
     logosContainer.append(avfLogo);
 
-    var unicefLogo = new ImageElement(src: 'assets/UNICEF-logo.svg')
```
lukechurch: We could Add Lark Social Impact :)

PR #249, opened 2020-03-17
Add a script to compute some tag analytics

marianamarasoiu:
This currently relies on a hard-coded list of tags as I didn't want to spam the output with e.g. age tags. If you'd prefer a different approach, plmk.

Thanks!


lukechurch:
@marianamarasoiu  thanks for putting this together- wondering a bit whether it's better to the push and pull Firebase <-> JSON to do the IO and a separate script just for processing...

Issue #242, opened 2020-03-15
Know which Git SHA is serving


lukechurch:
Because of propagation delays in Firebase, it would help if we could see on the UI which SHA was serving at a given point.

lukechurch:
@marianamarasoiu this might a good thing for Kenneth to help us with?

ktkization:
Here is my PL for this enhancement https://github.com/larksystems/nook/pull/258

marianamarasoiu:
Thanks @ktkization ! I'll assign this issue to me now to propagate the SHA to the UI.

ktkization:
[PL](https://github.com/larksystems/nook/pull/353) for this feature

ktkization:
@marianamarasoiu This can now be closed

marianamarasoiu:
thanks @ktkization, closing as fixed!

PR #258, opened 2020-03-18
Store the latest git commit hash on deployment

ktkization:
Re #242

### DIFF
--- a/deployment/deploy.sh
+++ b/deployment/deploy.sh
@@ -20,7 +20,8 @@ cd ..
 cp $PATH_TO_CONSTANTS_FILE public/assets/firebase_constants.json
 
 # Copy the latest commit sha1 hash on origin/master into the build folder
-LASTEST_COMMIT_HASH=$(git rev-parse origin/master)
+CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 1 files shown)


deployment/deploy.sh
```diff
@@ -19,6 +19,10 @@ cd ..
 # Copy the constants in the build folder
 cp $PATH_TO_CONSTANTS_FILE public/assets/firebase_constants.json
 
+# Copy the latest commit sha1 hash on origin/master into the build folder
+LASTEST_COMMIT_HASH=$(git rev-parse origin/master)
```
marianamarasoiu: I think it's possible (and likely) that the deploy script will be run from other places in the git tree rather than master. Is there a way of getting the current commit hash rather than the latest on master?

marianamarasoiu:
[please ignore the failing checks - they were an experiment in GitHub automation which doesn't seem to be working at the moment, it's nothing to do with your PR]

deployment/deploy.sh
ktkization: Yes, there is a way I think. Lemme refactor the code with the change

ktkization: Check the new changes. I've refactored it to get the hash from the current branch

Issue #256, opened 2020-03-18
Scrollbar doesn't work properly


lukechurch:
- Sometimes the message can appear as if it's at the bottom of the list, but if you scoll up, you can then scroll down again

lukechurch:
Fixed!

PR #259, opened 2020-03-18
improve scrolling in conversation list

danrubel:
This improves the scrolling in the lazily loaded conversations list by adding a single `_scrollPad` DOM element representing the not-yet-loaded DOM elements. The size of this DOM element is dynamically updated as new elements are loaded resulting in a scroll bar that closely approximates behavior if all of the elements were loaded. It also results in a small non-item white area at the end of the list so that you now can tell that you have indeed scrolled to the end and there are no more items.

Hopefully this will improve https://github.com/larksystems/nook/issues/256

Feel free to merge this PR if it proves helpful.

### DIFF
--- a/webapp/lib/lazy_list_view_model.dart
+++ b/webapp/lib/lazy_list_view_model.dart
@@ -13,7 +13,7 @@ class LazyListViewModel {
 
   /// The DOM element used to display the items
   /// and typically displaying only a subset of the items.
-  final _listView;
+  final DivElement _listView;
 
   /// The vertical scroll distance of the DOM elements in the [_listView] in pixels.
   int _scrollLength = 0;
@@ -21,6 +21,12 @@ class LazyListViewModel {
   /// The width of the [_listView] scrolling area in pixels
   /// or `null` if it has not been cached yet.
   int _scrollWidth = 10;
+
+  /// A DOM element used to pad the height of the scroll list
+  /// so that scrolling approximates the list size
+  /// without adding all of the individual DOM elements
+  final _scrollPad = DivElement();
+  static const _scollPadMinHeight = 30;
 
   /// When the width of the scrolling area changes,
   /// a delayed [Future] is created to recalculate [_scrollLength].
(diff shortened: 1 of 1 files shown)


lukechurch:
@marianamarasoiu could you take a first look at these - you're more familiar with this code than I am

webapp/lib/lazy_list_view_model.dart
```diff
@@ -80,18 +89,38 @@ class LazyListViewModel {
       // If the scroll area width changed, then recalculate the scroll length
       // because the item heights and thus the scroll length depends upon the scroll area width.
       _scrollLengthRecalc ??= new Future.delayed(const Duration(seconds: 2), () {
-        _scrollLength = _listView.children.fold(0, (len, element) => len + element.clientHeight);
+        _scrollLength = _listView.children.fold(0, (len, element)
+            => element == _scrollPad ? _scollPadMinHeight : len + element.clientHeight);
```
marianamarasoiu: instead of the _scrollPadMinHeight, shouldn't this be the actual height of the _scrollPad, potentially after the other elements have been resized?

webapp/lib/lazy_list_view_model.dart
```diff
@@ -13,15 +13,21 @@ class LazyListViewModel {
 
   /// The DOM element used to display the items
   /// and typically displaying only a subset of the items.
-  final _listView;
+  final DivElement _listView;
 
   /// The vertical scroll distance of the DOM elements in the [_listView] in pixels.
-  int _scrollLength = 0;
+  int _scrollLength = _scollPadMinHeight;
```
marianamarasoiu: typo _scoll.../_scroll...

danrubel: Oops. Good catch. Fixed.

webapp/lib/lazy_list_view_model.dart
danrubel: No, the _scrollLength is the vertical distance of the DOM elements that have been realized but does not include the items in the list that are not in the DOM. The `_scrollPadMinHeight` ensures a small blank space at the end of the list so that the users knows they have reached the end of the list.

danrubel:
@marianamarasoiu Thanks for the careful review!

danrubel:
@lukechurch I'm on the road today, so feel free to squash/merge this into master as you need.

webapp/lib/lazy_list_view_model.dart
marianamarasoiu: that makes sense now, thanks @danrubel !

lukechurch:
Thanks both!

Issue #257, opened 2020-03-18
Scrollbar jumps on Firestore update


lukechurch:
Sometimes if the message is in the middle, and there is a firebase update, the scrollbar jumps, losing the postion of where you were

lukechurch:
Thanks so much @danrubel this seems resolved

PR #260, opened 2020-03-18
update conversation list in place when processing updates from firebase

danrubel:
This modifies firebase update processing so that conversation summaries in the conversation list are updated in place rather than recomputing the entire list each time a new conversation update arrives.

Hopefully addresses https://github.com/larksystems/nook/issues/257

Feel free to squash/merge this if it is helpful.


lukechurch:
@marianamarasoiu could you take a first look at these - you're more familiar with this code than I am

danrubel:
@marianamarasoiu Thanks for the review!
@lukechurch I'm on the road today, so feel free to squash/merge this into master as you need.

PR #261, opened 2020-03-18
Compute daily analytics for nook conversations

marianamarasoiu:
The script has been quite substantially modified and repurposed, sorry for the substantial diff, it might be easier to just read it on its own rather than in diff form!

It now takes a backup of a Nook instance and a list of tags (likely exported from Coda), and computes the daily counts for each of those tags. Sample list included - the actual list lives in the project-specific repo.

Thanks!


tool/compute_tag_analytics.py
```diff
@@ -1,196 +1,173 @@
...
+            if "response_themes" not in date_metrics:
+                date_metrics["response_themes"] = {}
+            for tag in matched_tags:
+                date_metrics["response_themes"][tag_id_to_name[tag]] = True
+            conversation_metrics[date] = date_metrics
+
+        # add conversation metrics to overall metrics
+        for date, date_metrics in conversation_metrics.items():
+            total_date_metrics = daily_metrics.get(date, empty_daily_metrics(response_themes))
+            for key1 in date_metrics.keys():
```
lukechurch: This code is a bit surprising - I get what you're doing building something depending if it's a top level metric or a nested metric - but it might be worth a comment here?

marianamarasoiu: Thanks yeah, this is quite ugly code. I've added some more comments and also renamed the variables hoping that makes it a bit clearer.

marianamarasoiu:
thanks @lukechurch 
yes, I've validated the results, it's been quite the painful process!

I'll leave you to merge this when you think it's a good time and doesn't get in the way of the deployments

PR #262, opened 2020-03-20
Add needs reply analytics

marianamarasoiu:
[Note for merging: relies on #261 to be merged first, and then the destination branch of this PR will need changing to master]


lukechurch:
Thanks @marianamarasoiu  - this is great.

Probably also good to add a 'need-reply' and 'escalate' count metrics separately.

Once the covid response project is wrapped up, we'll need to work out how to make these project specific implementations - e.g. the Nutrition project has 'urgent' and 'sensitive' as separate high-priority tags. Please could you add a tracking issue for that refactoring?

marianamarasoiu:
@lukechurch Great idea, thanks - "needs-reply + escalate" count added. Please have another look!

marianamarasoiu:
thanks @lukechurch , merging now

Issue #253, opened 2020-03-18
Too much whitespace next to shortcut keys display


lukechurch:
please shrink whitespace to better support low res displays

danrubel:
@lukechurch Closing as this appears to have been fixed.

Issue #252, opened 2020-03-18
Send buttons over the top of text


lukechurch:
At the moment the Send (En) or Send (Swa) text is reserving whitespace next to the responses text. This impeeds usage on low res displays. It would be better if they hovered over the text about to be sent

danrubel:
@lukechurch Closing this as it appears to have been completed.

PR #265, opened 2020-03-20
Css updates

lukechurch:
Increase the density of replies in Nook


lukechurch:
Closes #253

lukechurch:
Closes #252

PR #277, opened 2020-03-23
Add script for measuring system metrics and pushing them to firebase

ktkization:


tool/Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-CoreDataModules = {git = "https://www.github.com/AfricasVoices/CoreDataModules", editable = true}
+CoreDataModules = {git = "https://www.github.com/AfricasVoices/CoreDataModules",editable = true}
```
marianamarasoiu: whitespace change, please revert

tool/metrics_monitor.py
```diff
@@ -0,0 +1,75 @@
...
+        metrics = {}
+
+        # record datetime
+        metrics['datetime'] = dt.datetime.now().isoformat() + 'Z'
+
+        # current cpu utlization
+        cpu_utilization = psutil.cpu_percent(interval=0.1)
+        metrics['cpu_percent'] = cpu_utilization
+
+        # cpu load over the last 1, 5 and 15 minutes in percentage
```
marianamarasoiu: Ah, right, the list of three values from the sample data makes sense now. Which leads me to wonder if we could be more transparent if we generated a map for this metric instead of a list, e.g. 
```{"1min": <value>, "5mins": <value>, "15mins": <value>}```
That way the view doesn't need to be implicitly aware of what each item in the list is.

File: tool/metrics_monitor.py [hunk 3]
```diff
@@ -0,0 +1,75 @@
...
+def initialize_firebase(CRYPTO_TOKEN_PATH):
+    global firebase_client
+    log.info("Setting up Firebase client")
+    firebase_cred = credentials.Certificate(CRYPTO_TOKEN_PATH)
+    firebase_admin.initialize_app(firebase_cred)
+    firebase_client = firestore.client()
+    log.info("Firebase client ready")
+
+
+def get_metrics(interval):
```
marianamarasoiu: rename to `get_and_publish_system_metrics` ?

tool/metrics_monitor.py
```diff
@@ -0,0 +1,75 @@
...
+import threading
+import sys
+import psutil
+import firebase_admin
+from firebase_admin import credentials
+from firebase_admin import firestore
+from core_data_modules.logging import Logger
+
+COLLECTION = 'pipeline_metrics'
+DEFAULT_INTERVAL = 3
```
marianamarasoiu: can you please add a comment here that this is in seconds?

tool/metrics_monitor.py
```diff
@@ -0,0 +1,75 @@
...
+        time.sleep(interval)
+
+
+def publish_metrics_to_firestore(metrics):
+    firebase_client.collection(COLLECTION).add(metrics)
+    log.info("Successfully published metrics to firebase {} collection".format(COLLECTION))
+
+
+def run_metric_monitor(interval=DEFAULT_INTERVAL):
+    initialize_firebase(sys.argv[1])
```
marianamarasoiu: We might need to make changes to this so that the firebase initialization happens outside this function call, but we can figure out exactly where and how when integrating this into the pipelines
cc @lukechurch

marianamarasoiu:
One more thought - I think `system_metrics_monitor.py` is more descriptive than jsut `metrics_monitor.py`, so can you please rename it to the former? thanks!

tool/Pipfile
ktkization: must have been the editor...I've reverted

tool/metrics_monitor.py
```diff
@@ -0,0 +1,75 @@
...
+    firebase_client = firestore.client()
+    log.info("Firebase client ready")
+
+
+def get_metrics(interval):
+    while True:
+        metrics = {}
+
+        # record datetime
+        metrics['datetime'] = dt.datetime.now().isoformat() + 'Z'
```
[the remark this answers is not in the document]
ktkization: Nice catch. Hadn't even noticed it was still not in UTC. I've refactored

tool/metrics_monitor.py
ktkization: This is a good idea. I've refactored the changes in

tool/metrics_monitor.py
ktkization: I agree. I've changed it

tool/metrics_monitor.py
ktkization: Added the comment. It's more descriptive now

tool/metrics_monitor.py
ktkization: Yeah... @lukechurch  has said he'd know where to refactor in this code into the pipeline

ktkization:
> One more thought - I think `system_metrics_monitor.py` is more descriptive than jsut `metrics_monitor.py`, so can you please rename it to the former? thanks!

I agree that this is more descriptive. I've renamed the file

tool/system_metrics_monitor.py
```diff
@@ -0,0 +1,80 @@
...
+            {
+                'available': memory_usage[1],
+                'used': memory_usage[3],
+                'percent': memory_usage[2],
+                'free': memory_usage[4]
+            }
+        )
+
+        # disk usage
+        disk_usage = psutil.disk_usage('/')
```
lukechurch: I think the standard setup we uses two disks, we'll need free-space on both

tool/system_metrics_monitor.py
```diff
@@ -0,0 +1,80 @@
...
+
+def publish_metrics_to_firestore(metrics):
+    firebase_client.collection(COLLECTION).add(metrics)
+    log.info("Successfully published metrics to firebase {} collection".format(COLLECTION))
+
+
+def run_system_metric_monitor(interval=DEFAULT_INTERVAL):
+    initialize_firebase(sys.argv[1])
+    runner = threading.Thread(target=get_and_publish_system_metrics, args=(interval,))
+    runner.start()
```
lukechurch: What happens if the thread running this throws an execption?

ktkization: It stops the execution. Do you think I should restart the thread again or what do you suggest?

tool/system_metrics_monitor.py
ktkization: I can modify this to show per disk. Do I do that?

tool/system_metrics_monitor.py
lukechurch: No - that's great. I'd rather it crash and we restart the whole script

tool/system_metrics_monitor.py
lukechurch: yes pls

ktkization: Done the refactor to show individual disk usages

marianamarasoiu:
Thanks @ktkization! merging now

Issue #279, opened 2020-03-23
Defect on messages being removed from list but not UI


lukechurch:
Repro:

Load UI
Apply filter (Need reply)
Mark some message as 'No need reply'
Wait for LS-Engine to propagate

Click on message in the list
- Error: invalid state for element will occur

PR #280, opened 2020-03-24
remove updated conversation from list if filter does not match

danrubel:
I failed to reproduce https://github.com/larksystems/nook/issues/279, but on reading the code it appears that a conversations would not be removed from the list. I think this is what's needed but not confident because I could not repo the problem or don't understand how to properly do so.

Feel free to merge this, use as part of some other PR, or toss this as you see fit.


danrubel:
The `_scrollPad.remove()` is not necessary but does not hurt.
`_updateScrollPad()` removes the scroll pad.

danrubel:
And thank you for fixing my fix!

PR #284, opened 2020-03-24
Remove item from DOM via removal from lazy list

marianamarasoiu:
Removing an item from DOM only resulted in the item not getting added back to the DOM if it needed to get back into the list. This now syncs the DOM list with the lazy list by removing the element from the DOM via the lazy list.

@danrubel In particular, does the update of the scrollPad in `removeItem` make sense? Should I be doing anything else when removing an item from the list?

Thanks!


danrubel:
`_updateScrollPad()` removes the scroll pad, so the additional call to `_scrollPad.remove()` is not necessary but does not hurt.

Issue #263, opened 2020-03-20
Need to parameterise which table name Nook is loading from


lukechurch:
We may well experience a scaling limit on loading conversations soon - LS-Engine can easily be modified to externalise to different tables in Firebase, if we can optionally parameterise the table name in the URL to nook it will make it easy to divide up work between the teams

lukechurch:
This will become critical w/c 23rd

marianamarasoiu:
@lukechurch @danrubel - can you confirm that conversations sharding has now fixed this issue?

danrubel:
@marianamarasoiu I implemented sharding one way, but @lukechurch had a better approach to sharding so I think it needs to be reworked. Also, we sharded some of the projects, but not others. I believe that this is still very much a work in progress and I don't know the priority of this compared with other issues.

marianamarasoiu:
Closing as out of date/duplicate of #316

PR #290, opened 2020-03-24
populate list of conversation lists

danrubel:
This PR loads shards from `/nook_conversation_shards` and populates the conversation list selection dropdown with what is found. Another step towards https://github.com/larksystems/nook/issues/263

* When the user selects a list, then the conversation list is populated  from `/nook_conversation_shards/<shard-doc-id>/conversations`

* If exactly one shard is found, then the conversation list dropdown is not shown
  and the conversations are loaded from that one shard

* If no shards are found, then the conversation list dropdown is not shown 
  and conversations are loaded from `nook_conversations`

Here are some screenshot of what the user sees if more than one shard found...

-----------------------

Screen Shot 2020-03-24 at 3 11 28 PM

-----------------------

Screen Shot 2020-03-24 at 3 13 36 PM

### DIFF
--- a/webapp/lib/model.yaml
+++ b/webapp/lib/model.yaml
@@ -1,3 +1,7 @@
+ConversationListShard:
+  firebaseCollectionName: 'nook_conversation_shards'
+  name: 'string'
+
 Conversation:
   firebaseCollectionName: 'nook_conversations'
   demographicsInfo: 'map string'
@@ -37,6 +41,8 @@ SuggestedReply:
   text: 'string'
   translation: 'publishable string'
   shortcut: 'string'
+  seq_no: 'int seqNumber'
+  category: 'string'
 
 Tag:
   firebaseDocId: 'string tagId'
--- a/webapp/lib/view.dart
+++ b/webapp/lib/view.dart
@@ -657,6 +657,7 @@ class ConversationListPanelView {
     conversationListPanel.append(_loadSpinner);
 
     _selectConversationListMessage = new DivElement()
+      ..classes.add('select-conversation-list-message')
       ..append(SpanElement()..text = "Select a conversation list above")
       ..hidden = true;
     conversationListPanel.append(_selectConversationListMessage);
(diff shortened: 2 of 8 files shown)


webapp/lib/model.yaml
```diff
@@ -37,6 +41,8 @@ SuggestedReply:
   text: 'string'
   translation: 'publishable string'
   shortcut: 'string'
+  seq_no: 'int seqNumber'
+  category: 'string'
```
marianamarasoiu: Thanks for updating these as well!

webapp/lib/view.dart
```diff
@@ -575,10 +574,45 @@ class ConversationListSelectHeader {
...
         ..classes.add('conversation-list-select')
-        // TODO dynamically populate this from firebase "tables/conversation-lists"
-        ..add(OptionElement(data: 'nook_conversations'), null));
+        ..add(OptionElement(data: '... loading conversation lists ...'), null)
+        ..onChange.listen(shardSelected));
+  }
+
+  void updateConversationLists(List<ConversationListShard> shards) {
+    // TODO Consider displaying summary information about each shard... e.g. # of unread conversations
+    if (_shards == null) {
```
marianamarasoiu: Doesn't this mean that if the list of shards changes in firebase, it won't update in the UI without a page refresh?

marianamarasoiu: I think ideally we would like it to update in the conversation list selector automatically, and if possible, to keep the user on the shard they're on etc. but this is probably something to follow up in a future PR

danrubel: @marianamarasoiu There are two ways that the shards can change

* name or status information in the shard description changes in which case, yes, we should refresh the conversation list

* a new shard is added which means that conversations have moved to different shards. In this situation we should probably force refresh the page so that it displays reality again. We need to guard against a client with out of date shards unknowingly sending invalid changes to the backend

I am not clear or sure about the above so I left it for a future PR.

Issue #286, opened 2020-03-24
UI customisation per user


lukechurch:
Need to add customisation per user

- Tags UI visible (default: true)
- Send custom message enabled (default: false)
- Send multi-message enabled (default: false)
- Keyboard shortcuts enabled (default: false)

For now, configuration stored in a firebase table, per user

lukechurch:
please can we separate the keyboard shortcuts for sending messages from the ones for setting tags?

PR #291, opened 2020-03-24
Makes several UI features customizable (turn on / off) via a firebase users table

marianamarasoiu:
Re #286 - only 3/4, toggling tag panel visibility is lower priority and requires a bit more CSS work than just marking it as hidden, will follow up in a next PR.

@lukechurch At the moment, these get set when the user first loads Nook and don't get updated during the session (as I didn't expect they would need to), but if they should get updated throughout, plmk and I'll add that in.

Thank you!

### DIFF
--- a/webapp/lib/platform.dart
+++ b/webapp/lib/platform.dart
@@ -42,6 +42,33 @@ init() async {
     _docStorage = FirebaseDocStorage(firebase.firestore());
     _pubsubInstance = new PubSubClient(platform_constants.publishUrl, user);
     controller.command(controller.UIAction.userSignedIn, new controller.UserData(user.displayName, user.email, photoURL));
+
+    // Get user features
+    firebase.firestore().collection('users').get().then((querySnapshot) {
+      var documents = querySnapshot.docs;
+      var userFeatures = documents.where((docSnapshot) => docSnapshot.id == user.email);
+      if (userFeatures.isNotEmpty) {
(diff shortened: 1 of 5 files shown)


webapp/lib/platform.dart
```diff
@@ -42,6 +42,33 @@ init() async {
     _docStorage = FirebaseDocStorage(firebase.firestore());
     _pubsubInstance = new PubSubClient(platform_constants.publishUrl, user);
     controller.command(controller.UIAction.userSignedIn, new controller.UserData(user.displayName, user.email, photoURL));
+
+    // Get user features
```
lukechurch: This code is difficult for me to read - as this callback is actually executed below the call below - could we swap the order?

I wonder if we could also refactor it into a support method that that takes a data object and applies to cull the duplication a bit?

lukechurch: Follow on PR is fine :)

lukechurch:
@marianamarasoiu I think it would be great if these could be dymamic actually - I think there may well be times when we need to turn features in the UI on and off during the course of a session

webapp/lib/platform.dart
marianamarasoiu: Thanks @lukechurch  I agree it was a bit unwieldy, and an accidental mis-indent was also making it harder to read.

I've refactored setting the flags into a separate function, and while doing so it occurred to me that it would be a good idea to always use the default settings from Firebase over the initial UI-set settings rather than only when there are no specific user preferences, in order to support single flag enables per user, if that makes sense?

Anyway, please have another look at this part of the code (L45-75) and let me know if you have any other concerns. Thanks!

PR #292, opened 2020-03-25
Provide help information on arguments needed to run script.

### LINKED ISSUE #288: system_monitor script doesn't provide help text
```
lukes-mbp:tool lukechurch$ pipenv run python system_metrics_monitor.py 
Traceback (most recent call last):
  File "system_metrics_monitor.py", line 83, in <module>
    run_system_metric_monitor()
  File "system_metrics_monitor.py", line 78, in run_system_metric_monitor
    initialize_firebase(sys.argv[1])
IndexError: list index out of range
```

@ktkization  Please could you fix so that if called without argument it provides help text explaining how it should be run?

ktkization:
Fixes #288

### DIFF
--- a/tool/system_metrics_monitor.py
+++ b/tool/system_metrics_monitor.py
@@ -77,10 +77,10 @@ def publish_metrics_to_firestore(metrics):
 
 def run_system_metric_monitor(interval=DEFAULT_INTERVAL):
     parser = argparse.ArgumentParser(description='Retrieve system metrics i.e cpu utilization, memory & disk usage')
-    parser.add_argument('cert', metavar='cert', type=str, help='path to Firebase JSON certificate keyfile')
+    parser.add_argument("crypto_token_file", type=str, help="path to Firebase crypto token file")
     args = parser.parse_args()
 
-    initialize_firebase(args.cert)
+    initialize_firebase(args.crypto_token_file)
     runner = threading.Thread(target=get_and_publish_system_metrics, args=(interval,))
     runner.start()
 
--- a/tool/system_metrics_monitor.py
+++ b/tool/system_metrics_monitor.py
@@ -2,6 +2,7 @@ import time
 import datetime as dt
 import threading
 import sys
+import argparse
 import psutil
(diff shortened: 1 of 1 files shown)


ktkization:
Can you please add yourselves as reviewers @marianamarasoiu

tool/system_metrics_monitor.py
```diff
@@ -75,7 +76,11 @@ def publish_metrics_to_firestore(metrics):
 
 
 def run_system_metric_monitor(interval=DEFAULT_INTERVAL):
-    initialize_firebase(sys.argv[1])
+    parser = argparse.ArgumentParser(description='Retrieve system metrics i.e cpu utilization, memory & disk usage')
+    parser.add_argument('cert', metavar='cert', type=str, help='path to Firebase JSON certificate keyfile')
```
marianamarasoiu: Can you use the same naming scheme we've been using throughout the other scripts? ie. `crypto_token` instead of `cert`/certificate keyfile? e.g. something like in [/firebase_to_json.py](https://github.com/larksystems/nook/blob/master/tool/firebase_to_json.py)

I also don't think you need to set the `metavar` if it's the same as the name?

tool/system_metrics_monitor.py
```diff
@@ -75,7 +76,11 @@ def publish_metrics_to_firestore(metrics):
 
 
 def run_system_metric_monitor(interval=DEFAULT_INTERVAL):
-    initialize_firebase(sys.argv[1])
+    parser = argparse.ArgumentParser(description='Retrieve system metrics i.e cpu utilization, memory & disk usage')
```
marianamarasoiu: Thanks for adding in the description, we don't do this often enough!

tool/system_metrics_monitor.py
ktkization: I've changed it to use the same scheme

tool/system_metrics_monitor.py
ktkization: cool :)

PR #295, opened 2020-03-26
Listen to user configuration of UI features changes from firebase

### LINKED ISSUE #286: UI customisation per user

marianamarasoiu:
Closes #286 

Follows up from #291 by replacing the single read of user configuration from firebase into a listener for any changes, and adding in the ability to hide/show the tag panel.

Thanks!

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -311,15 +311,15 @@ void initUI() {
 
 /// Sets user customization flags from the data map
 /// If a flag is not set in the data map, it defaults to the existing values
-void applyConfiguration(model.UserConfiguration userConfiguration) {
-  view.replyPanelView.showShortcuts(userConfiguration.keyboardShortcutsEnabled ?? false);
-  view.tagPanelView.showShortcuts(userConfiguration.keyboardShortcutsEnabled ?? false);
+void applyConfiguration(model.UserConfiguration config) {
+  view.replyPanelView.showShortcuts(config.keyboardShortcutsEnabled ?? false);
+  view.tagPanelView.showShortcuts(config.keyboardShortcutsEnabled ?? false);
 
-  view.conversationPanelView.showCustomMessageBox(userConfiguration.sendCustomMessagesEnabled ?? false);
+  view.conversationPanelView.showCustomMessageBox(config.sendCustomMessagesEnabled ?? false);
 
-  view.conversationListPanelView.showConversationSelectCheckboxes(userConfiguration.sendMultiMessageEnabled ?? false);
+  view.conversationListPanelView.showConversationSelectCheckboxes(config.sendMultiMessageEnabled ?? false);
 
-  // TODO(mariana): Implement hide/show for the tag panel
+  view.showTagPanel(config.tagPanelVisibility ?? false);
 }
 
 void conversationListSelected(String conversationListRoot) {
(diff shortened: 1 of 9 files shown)


webapp/lib/controller.dart
```diff
@@ -308,6 +294,32 @@ void initUI() {
...
     });
+
+  platform.listenForUserConfigurations((userConfigurations) {
+    var defaultConfig = userConfigurations.singleWhere((c) => c.docId == 'default', orElse: () => null);
+    var userConfig = userConfigurations.singleWhere((c) => c.docId == signedInUser.userEmail, orElse: () => null);
+    if (defaultConfig == null && userConfig == null) {
+      return;
+    }
+    defaultUserConfig = defaultConfig ?? defaultUserConfig;
+    currentUserConfig = userConfig ?? currentUserConfig;
```
danrubel: What if `defaultConfig` is not `null` and `currentConfig` is `null` or vise-versa?
Are `defaultUserConfig` and `currentUserConfig` defined as non-`null` somewhere else prior to this?
I'm guessing they are and I'm just not seeing it.

webapp/lib/model.dart
```diff
@@ -6,6 +6,16 @@ export 'model.g.dart' hide
   MessageDirection_fromStringOverride,
   TagType_fromStringOverride;
 
+extension UserConfigurationUtil on g.UserConfiguration {
+  g.UserConfiguration operator +(g.UserConfiguration other) =>
```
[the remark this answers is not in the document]
marianamarasoiu: This is a very good point, I hadn't thought about maintaining commutativity. I like your solution so I've rewritten the code to use `applyDefaults` instead of `+`, thank you!

webapp/lib/controller.dart
marianamarasoiu: good catch, it's true I'm not initialising `defaultUserConfig` and `currentUserConfig` anywhere, and they should be, otherwise on the following line, the `currentConfig` can switch to being null, which shouldn't happen. I've fixed this now, thank you!

webapp/lib/model.g.dart
```diff
@@ -427,6 +427,44 @@ class SystemMessage {
 }
 typedef void SystemMessageCollectionListener(List<SystemMessage> changes);
 
+class UserConfiguration {
+  static const collectionName = 'users';
```
lukechurch: consider COLLECTION_NAME

marianamarasoiu: Thanks for flagging it, I think it's a good idea but probably best not to do it in this PR for consistency.
The camel case version is generated by the model generator, and since it applies to multiple classes in this file, I think it would be best changing it there first, and follow up with a PR for the whole `model.g.dart` afterwards rather than doing it just locally here.

marianamarasoiu:
Thank you both for the review, I've needed to make a few small changes to the configuration algorithm following the merge of #300 for listening to add/modify/remove firebase changes. @danrubel could I ask you to have another quick look at the code in `controller.dart` when you you get the chance before I merge? thanks!

webapp/lib/model.g.dart
danrubel: @marianamarasoiu I need to update codegen to handle another change, so I'll tackle this at the same time.

marianamarasoiu: oh thanks @danrubel that would be great!

PR #300, opened 2020-03-27
Separate firebase doc changes into added/modified/removed

marianamarasoiu:
This propagates doc removals from firebase into the UI.

At the moment it's not working well for conversations (which get removed from the internal model but not from the UI) and conversations shards. Will follow up with separate PRs to fix conversations and conversation shards.


lukechurch:
Thanks @marianamarasoiu

marianamarasoiu:
thank you both for the review, merging now!

PR #302, opened 2020-03-29
update firebase root keys for json download

danrubel:
A change for if/when we move to sharded conversations.

Feel free to merge this when the time is right.

### DIFF
--- a/tool/firebase_root_keys.py
+++ b/tool/firebase_root_keys.py
@@ -1,6 +1,7 @@
 root_keys = [
     "conversationTags",
     "messageTags",
+    "nook_conversations",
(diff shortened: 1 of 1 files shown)


marianamarasoiu:
@danrubel it looks like we're going to want to maintain `nook_conversations` and `nook_conversation_shards` in parallel for a while. Can you change this PR to add the shards root key rather than replace the other collection? I think it'll be ok to merge after that. Thanks!

danrubel:
Good point. Fixed.

marianamarasoiu:
great, thanks @danrubel , merging now..

PR #303, opened 2020-03-29
Integrate New Logger

lukechurch:
This is an example of what integrating the new logger will look like, fairly minimal changes as you can see. I'll roll this out across the rest of the tools if you're happy

### DIFF
--- a/tool/firebase_to_json.py
+++ b/tool/firebase_to_json.py
@@ -103,7 +103,7 @@ if __name__ == '__main__':
     data = {}
 
     short_id = tool_utils.short_id()
-    log.audit(f"firebase_to_json: JobID {short_id}")
+    log.audit(f"firebase_to_json: JobID ({short_id}), keys to download {json.dumps(root_keys_to_export)}")
     for key in root_keys_to_export:
         data[key] = import_data_for_firestore_col_root(key)
 
--- a/tool/firebase_to_json.py
+++ b/tool/firebase_to_json.py
@@ -4,19 +4,22 @@
 import firebase_admin
 from firebase_admin import credentials
 from firebase_admin import firestore
-from core_data_modules.logging import Logger
+from katikati_pylib.logging import logging
+
 from firebase_root_keys import root_keys
+import tool_utils
 import time
 import argparse
 import json
 import sys
 
-log = Logger(__name__)
-
+log = None
 firebase_client = None
 
 def init(CRYPTO_TOKEN_PATH):
     global firebase_client
+    global log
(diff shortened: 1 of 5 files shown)


tool/tool_utils.py
```diff
@@ -0,0 +1,4 @@
+import uuid
+
+def short_id():
+    return str(uuid.uuid4()).split("-")[0]
```
danrubel: Why only include the short id in the audit log?
If we're including a uuid why not the entire uuid4?
I don't mean you should change this... just me learning the reasons behind the decision....

tool/firebase_to_json.py
```diff
@@ -99,9 +102,12 @@ def _usage_and_exit(error_message):
 
     data = {}
 
+    short_id = tool_utils.short_id()
+    log.audit(f"firebase_to_json: JobID {short_id}")
     for key in root_keys_to_export:
```
danrubel: Do we want to `log.notify` the keys that are exported?

tool/json_to_firebase.py
```diff
@@ -68,6 +71,8 @@ def _usage_and_exit(error_message):
 
     init(CRYPTO_TOKEN_PATH)
 
+    short_id = tool_utils.short_id()
+    log.audit(f"json_to_firebase: JobID {short_id}")
     with open(INPUT_PATH, 'r') as f:
         data_dict = json.load(f)
```
danrubel: Do we want to `log.notify` the top level keys being uploaded to firebase?

tool/tool_utils.py
lukechurch: Sure

The purpose of this is to enable us to link IDs together using grep, the short IDs are easier to work with and don't have the same impact on making the log unreadable. (c.f. the way in which GitHub's short shas work)

There is other metadata that can be used in the event of a collision - e.g. the name of the machine from which the request originated - so the risk of not being able to determine them is pretty negligible. It may turn out that we need to do machine reading of these logs rather than grep etc, in which case I suggest we enrich the API and use the full UUID

tool/firebase_to_json.py
lukechurch: Good idea

tool/json_to_firebase.py
lukechurch: Good Idea, will adjust, thanks

tool/firebase_to_json.py
lukechurch: Done

tool/json_to_firebase.py
lukechurch: Done

PR #304, opened 2020-03-29
Add conversations to the list in one go rather than one by one

marianamarasoiu:
This reduces load time delays due to DOM render and reflow from about 7s to 250ms on my machine for 2k conversations.

### DIFF
--- a/webapp/lib/lazy_list_view_model.dart
+++ b/webapp/lib/lazy_list_view_model.dart
@@ -44,7 +44,7 @@ class LazyListViewModel {
       position = 0;
     }
     _items.insert(position, item);
-    if (position < _listView.children.length) {
+    if (position < _listView.children.length - 1) {
       // Insert the item's element into the cached/visible DOM elements
       Node refChild = _listView.children[position];
       _listView.insertBefore(item.element, refChild);
@@ -52,6 +52,11 @@ class LazyListViewModel {
     } else {
       _updateCachedElements();
     }
+  }
+
+  void appendItems(Iterable<LazyListViewItem> items) {
+    _items.addAll(items);
+    _updateCachedElements();
   }
 
   void clearItems() {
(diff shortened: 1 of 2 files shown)


webapp/lib/lazy_list_view_model.dart
```diff
@@ -44,7 +44,7 @@ class LazyListViewModel {
       position = 0;
     }
     _items.insert(position, item);
-    if (position < _listView.children.length) {
+    if (position < _listView.children.length - 1) {
```
marianamarasoiu: This is because the `_listView.children` also includes the `_scollPad`, so it's not the same length as `_items`. Without this change, the if clause was always positive, and so new items were always being added to the DOM rather than lazily added.

danrubel: Good catch! Thanks!

webapp/lib/view.dart
```diff
@@ -730,28 +730,43 @@ class ConversationListPanelView {
...
+      conversationsToAdd.add(summary);
+    }
+    _conversationList.appendItems(conversationsToAdd);
+    for (var conversation in conversationsToAdd) {
+      _phoneToConversations[conversation.deidentifiedPhoneNumber] = conversation;
     }
+    _conversationPanelTitle.text = '${_phoneToConversations.length} conversations';
   }
 
   void addOrUpdateConversation(Conversation conversation) {
```
danrubel: Is this method used anymore?
Can it be removed now that you have improved the `updateConversationList` above?

marianamarasoiu: It's not used anymore, and I considered removing it, but I think I'd like to keep it around for a bit longer in case we end up needing it. I'll set a timer on it for a couple of months from now, and if we haven't used it by then, we can remove it.

PR #311, opened 2020-04-01
Update analytics code to consider sharded conversations

marianamarasoiu:


tool/compute_tag_analytics.py
```diff
@@ -12,6 +12,7 @@
 log = Logger(__name__)
 
 CONVERSATIONS_COLLECTION_KEY = 'nook_conversations'
+CONVERSATIONS_SHARDS_COLLECTION_KEY = 'nook_conversation_shards'
```
danrubel, line 15: I'd recommend CONVERSATION_SHARDS_COLLECTION_KEY (conversation is singular) to match the actual key

tool/compute_tag_analytics.py
```diff
@@ -279,7 +285,16 @@ def _usage_and_exit(error_message):
     with open(NOOK_EXPORT, mode="r") as nook_export_file:
         nook_export_data = json.load(nook_export_file)
     conversation_tags = nook_export_data[CONVERSATION_TAGS_COLLECTION_KEY]
-    nook_conversations = nook_export_data[CONVERSATIONS_COLLECTION_KEY]
+
+    # If the data is nook_conversations, use it, otherwise try to read from nook_conversation_shards
+    if CONVERSATIONS_COLLECTION_KEY in nook_export_data:
+        nook_conversations = nook_export_data[CONVERSATIONS_COLLECTION_KEY]
+    elif CONVERSATIONS_SHARDS_COLLECTION_KEY in nook_export_data:
```
danrubel, line 292: I recommend putting this if statement before `if CONVERSATIONS_COLLECTION_KEY` because if there are shards, they should probably take precedence over stale data in `nook_conversations`.

marianamarasoiu: That's a good point, thanks, and I see that the UI is also considering shards first, so good idea to do the same here. Fixed now.

tool/compute_tag_analytics.py
marianamarasoiu: Thanks for the catch, fixed!

tool/compute_tag_analytics.py
```diff
@@ -279,7 +285,16 @@ def _usage_and_exit(error_message):
     with open(NOOK_EXPORT, mode="r") as nook_export_file:
         nook_export_data = json.load(nook_export_file)
     conversation_tags = nook_export_data[CONVERSATION_TAGS_COLLECTION_KEY]
-    nook_conversations = nook_export_data[CONVERSATIONS_COLLECTION_KEY]
+
+    # If the data is nook_conversations, use it, otherwise try to read from nook_conversation_shards
```
lukechurch, line 289: nit: This comment doesn't match the logic below which tests nook_conversation_shards first

tool/compute_tag_analytics.py
lukechurch: Ahh, I see this is why the comment was wrong - please update to match new logic

lukechurch:
Thanks both, LGTM

tool/compute_tag_analytics.py
marianamarasoiu: thanks for the catch, done!

PR #314, opened 2020-04-02
Add project name to tag analytics script

marianamarasoiu:


tool/firebase_root_keys.py
```diff
@@ -2,6 +2,7 @@
     "conversationTags",
     "messageTags",
     "nook_conversations",
+    "nook_conversation_shards",
```
marianamarasoiu, line 5: I'm not sure why this appears here as a new addition, as I didn't add it in any of my commits, and when I browse nook master it's there anyway. It seems that git is slightly broken?

marianamarasoiu:
Merging now to test, @danrubel - feel free to have a look and comment post-merge, thanks!

PR #325, opened 2020-04-06
Update checkboxes logic

marianamarasoiu:
With this PR, the conversation list checkboxes will be shown all the time if the `send_multi_message_enabled` flag is set to `true`, as opposed to only when the select all checkbox is ticked. This is to support selecting messages one by one, starting from no selected messages.

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -46,8 +46,8 @@ enum UIAction {
   keyPressed,
   addNewSuggestedReply,
   addNewTag,
-  enableMultiSelectMode,
-  disableMultiSelectMode,
+  selectAllConversations,
+  deselectAllConversations,
   updateSystemMessages,
   updateSuggestedRepliesCategory,
   hideAgeTags
@@ -193,7 +193,6 @@ model.UserConfiguration currentUserConfig;
 model.UserConfiguration currentConfig;
 
 bool hideDemogsTags;
-bool multiSelectMode;
 
 void init() async {
   defaultUserConfig = currentConfig = baseUserConfiguration;
@@ -210,7 +209,6 @@ void initUI() {
   conversationTags = [];
   messageTags = [];
   selectedConversations = [];
-  multiSelectMode = false;
   activeConversation = null;
   selectedSuggestedRepliesCategory = '';
   hideDemogsTags = true;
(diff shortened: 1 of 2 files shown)


webapp/lib/controller.dart
```diff
@@ -350,7 +348,11 @@ void applyConfiguration(model.UserConfiguration config) {
 
   view.conversationPanelView.showCustomMessageBox(config.sendCustomMessagesEnabled ?? false);
 
-  view.conversationListPanelView.showConversationSelectCheckboxes(config.sendMultiMessageEnabled ?? false);
+  if (config.sendMultiMessageEnabled) {
```
danrubel, line 351: Would it make more sense to rename this config feature `mulitMessageActionEnabled` or something like that?
Or does it only enable sending multiple messages?

webapp/lib/controller.dart
```diff
@@ -587,19 +586,19 @@ void command(UIAction action, Data data) {
...
-          if (!conversation.unread) {
-            markedConversations.add(conversation);
-            view.conversationListPanelView.markConversationUnread(conversation.docId);
-          }
-        }
-        platform.updateUnread(markedConversations, true).catchError(showAndLogError);
-      } else {
+      if (!currentConfig.sendMultiMessageEnabled || selectedConversations.isEmpty) {
         view.conversationListPanelView.markConversationUnread(activeConversation.docId);
         platform.updateUnread([activeConversation], true).catchError(showAndLogError);
```
danrubel, line 591: Would it make sense for `selectedConversations` to have the following characteristics?
* always non-null
* always contain `activeConversation` if `activeConversation != null`
* always be empty if `activeConversation == null`

If this were true, then several action handlers like this one could be simplified.

webapp/lib/view.dart
```diff
@@ -917,10 +914,10 @@ class ConversationSummary with LazyListViewItem {
       ..classes.add('conversation-selector')
       ..title = 'Select conversation'
       ..checked = _checked
-      ..style.visibility = 'hidden'
+      ..hidden = _checkboxVisible
```
danrubel, line 917: hidden = visible ?
Based on the code below, perhaps this should be `..hidden = false` ?
And maybe the field initialization above should be `bool _checkboxVisible = true;` to match ?

marianamarasoiu: Thanks @danrubel, yeah, this was a bit messy. Switched to a `_checkboxHidden` variable which makes things more straighforward.

webapp/lib/controller.dart
marianamarasoiu: Thanks @danrubel I'll need to have a think in more detail about this as the separation between them seems helpful - in multiConversationSelect mode, the active conversation is just an indicator for the conversation you're seeing in the second panel which supports a workflow where the user goes through the list, reads each conversation, and only selects it if they want to add it to a bulk reply list. Always considering `activeConversation` as part of the `selectedConversations` would require changing that workflow or doing something else in the UI to support it.

webapp/lib/controller.dart
marianamarasoiu: yes, I think that would make sense - it also enables tagging multiple conversations and also marking multiple conversations as unread. I'll follow up with a PR to change this. thanks for the suggestion!

PR #327, opened 2020-04-06
remove duplicate deploy scripts

danrubel:
This removes duplicate deploy scripts in favor of `deploy_webapp_with_metadata.sh` which is now renamed to `deploy_webapp.sh`


lukechurch:
@danrubel - could you put in some code verifying that the crypto-token associated with the same project as the firebase constants? Otherwise we could be thinking we're deploying to one place and actually pushing the data to another

PR #329, opened 2020-04-07
message status and logging improvements

danrubel:
If send reply or send multi reply fails, then 
* a snackbar is displayed indicating the failure
* the message text is displayed with red strike-through

Also, publish logging has been enhanced to better collate http responses

image

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -889,19 +889,20 @@ void sendReply(model.SuggestedReply reply, model.Conversation conversation) asyn
     ..translation = reply.translation
     ..tagIds = [];
   conversation.messages.add(newMessage);
-  view.conversationPanelView.addMessage(
-    new view.MessageView(
+  var newMessageView = new view.MessageView(
       newMessage.text,
       newMessage.datetime,
       conversation.docId,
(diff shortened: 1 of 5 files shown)


webapp/lib/controller.dart
```diff
@@ -881,47 +881,64 @@ void updateViewForConversation(model.Conversation conversation) {
...
       conversation.docId,
       conversation.messages.indexOf(newMessage),
       translation: newMessage.translation,
-      incoming: false)
-  );
-  platform.sendMessage(conversation.docId, reply.text);
+      incoming: false);
+  view.conversationPanelView.addMessage(newMessageView);
+  try {
+    await platform.sendMessage(conversation.docId, reply.text);
```
marianamarasoiu, line 901: What are your thoughts to adding an (optional) `onError` parameter to `sendMessage`? I could be persuaded both ways, but I'm wondering if it would be nicer to catch the `PubSubException` at platform level, rather than here?

danrubel: @marianamarasoiu Good suggestion. I think it looks better with `onError` as you suggest
and gives us a central place to make error handling changes... https://github.com/larksystems/nook/pull/331

PR #330, opened 2020-04-07
Add extra logging of UI commands and to send reply functions

marianamarasoiu:


marianamarasoiu:
I needed to rebase my PR on top of yours and so went on and removed the await/async for the send reply methods - can you please have another look? Thanks!

I also needed to change the text of the last log from "sent" to "queued for sending" as initially I didn't realise that it `platform.send` was asynchronous. It did make me consider keeping the `await` in, but in the end we wouldn't know if it's been successful or not, only the `onError` does, so the logs wouldn't have been much clearer either.

PR #331, opened 2020-04-07
refactor send reply to have onError param

danrubel:
This refactors `sendReply` and `sendMultiReply` to use a new `onError` parameter to handle send message exceptions.

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -897,16 +897,11 @@ void sendReply(model.SuggestedReply reply, model.Conversation conversation) asyn
       translation: newMessage.translation,
       incoming: false);
   view.conversationPanelView.addMessage(newMessageView);
-  try {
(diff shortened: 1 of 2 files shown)


webapp/lib/controller.dart
```diff
@@ -897,16 +897,11 @@ void sendReply(model.SuggestedReply reply, model.Conversation conversation) asyn
       translation: newMessage.translation,
       incoming: false);
   view.conversationPanelView.addMessage(newMessageView);
-  try {
-    await platform.sendMessage(conversation.docId, reply.text);
-  } catch (_) {
+  await platform.sendMessage(conversation.docId, reply.text, onError: (_) {
```
marianamarasoiu, line 900: Do we still need the `await`?

danrubel: If we omit `await` then it will not wait for the sendMessage future to complete and the exception will happen after we have exited the try/catch block.

PR #333, opened 2020-04-09
Add uptime monitoring every min and every 5 secs if minute ping fails

marianamarasoiu:
This also adds a snackbar/toast that could be triggered from the platform. I initially did this so I can trigger the snackbar to show failures, but since it's so intrusive, I ended up using the banner notification instead.

### DIFF
--- a/webapp/lib/platform.dart
+++ b/webapp/lib/platform.dart
@@ -66,6 +66,8 @@ void initUptimeMonitoring() {
       };
       _uptimePubSubInstance.publish(platform_constants.statuszTopic, payload).then(
         (_) {
+          log.success('Uptime ping ${t.tick}.${tt.tick} successful');
+
           // Add success to the three pings queue
           lastThreePingsQueue.add(true);
           lastThreePingsQueue.removeAt(0);
@@ -116,6 +118,8 @@ void initUptimeMonitoring() {
     };
     _uptimePubSubInstance.publish(platform_constants.statuszTopic, payload).then(
       (_) {
+        log.success('Uptime ping ${t.tick} successful');
+
         // Cancel the previous 5 sec timer if it's still going.
         fiveSecTimer?.cancel();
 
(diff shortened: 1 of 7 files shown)


webapp/lib/pubsub.dart
```diff
@@ -20,8 +20,8 @@ class PubSubClient extends DocPubSubUpdate {
 
   PubSubClient(this.publishUrl, this.user);
 
-  /// Publish the specified exception.
-  /// Callers should catch and handle IOException.
```
danrubel, line 24: `PubSubException` is only one of the exceptions that can be thrown here.
Perhaps `Callers should catch and handle PubSubException and IOException` instead?

marianamarasoiu: Good point, thanks. But I don't think IO exception exists for web code? IIRC, this comment was left out of date from when we accidentally imported `dart:io` and had `PubSubException` inherit from `IOException`

Looking at the `http` library, it looks like the only exception they define is `ClientException`, so I've added that instead. If there's another exception that would be good to list here, please let me know.

webapp/lib/platform.dart
```diff
@@ -42,6 +43,129 @@ init() async {
...
+    });
+  };
+
+  new Timer.periodic(oneMinute, (Timer t) {
+    Map payload = {
+      'ping': '${t.tick}',
+      'lastUserActivity': controller.lastUserActivity.toIso8601String()
+    };
+    _uptimePubSubInstance.publish(platform_constants.statuszTopic, payload).then(
+      (_) {
```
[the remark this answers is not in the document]
marianamarasoiu: Good idea, thanks, done!

webapp/lib/pubsub.dart
danrubel: @marianamarasoiu Ah yes. That sounds right. Thanks!

marianamarasoiu:
Thanks both for the review! merging now..

PR #335, opened 2020-04-09
Add snackbar UI command

marianamarasoiu:
This is so the snackbar/toast could be triggered from the platform.

I initially did this to support #333 , but ended up using the banner notification instead.


marianamarasoiu:
Closing this rather than merging - I thought I could rebase #333 against this, but github doesn't do a simple diff, it instead uses the commits, which are different, so the changes here still appear in the other PR. Sorry for the confusion!

PR #337, opened 2020-04-10
Compare project ids & fix relative path issue

marianamarasoiu:
This introduces an upfront check between the firebase constants file and the crypto token to ensure that they both point to the same project.

It also fixes an issue where the script args needed to be passed either as absolute paths, or needed to be under the project repo root (because of the `cd "$(dirname "$0")"/..` command).

### DIFF
--- a/tool/deploy_webapp.sh
+++ b/tool/deploy_webapp.sh
@@ -5,36 +5,69 @@ if [ $# -ne 2 ]; then
     exit 1
 fi
 
-########## cd to the nook project directory and get the absolute path
-
-cd "$(dirname "$0")"/..
-PROJDIR="$(pwd)"
+WORK_DIR="$(pwd)"
 
 ########## validate argument(s)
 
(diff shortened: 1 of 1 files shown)


tool/deploy_webapp.sh
```diff
@@ -5,37 +5,70 @@ if [ $# -ne 2 ]; then
...
+  echo "the two project ids must match, check that you're deploying to the correct project"
   exit 1
 fi
 
+########## cd to the nook project directory and get the absolute path
+
+cd "$WORK_DIR"
+
+cd "$(dirname "$0")"/..
+PROJECT_DIR="$(pwd)"
```
danrubel, line 57: Should we rename this "NOOK_DIR" to keep the same env var naming conventions as in the KK project scripts?

marianamarasoiu: Good idea, done!

PR #340, opened 2020-04-14
Add cd to work dir after getting full paths

marianamarasoiu:
Missed this when testing the previous change - which caused the script to fail for relative crypto token paths.


marianamarasoiu:
thanks @danrubel!

PR #341, opened 2020-04-15
Add an outbox feature for storing messages before being sent through

marianamarasoiu:
@danrubel , @lukechurch 
This is still a WIP, and a substantial change that will need a good bit of testing before being used, so I wanted to give you some early visibility of what's going on. Feel free to take it for a spin, and let me know if you find any bugs, or have any suggestions for improvement.

Todolist before this is ready:

- [ ] Warn against leaving the UI with messages outstanding in the outbox
- [ ] Warn if new messages came in a conversation since messages have been added into the outbox
- [ ] Better handling of failed messages - e.g. offer to resend
- [ ] User customization - send directly without outbox, only add messages to outbox without send
- [ ] Test against a running instance of the katikati infrastructure
- [ ] Outbox messages are being tracked only locally, but should (?) be sent to the platform. Figure out how that's going to be shared amongst multiple users (cc @lukechurch)


webapp/lib/controller.dart
```diff
@@ -597,17 +634,72 @@ void command(UIAction action, Data data) {
...
+          break;
+      }
+      break;
+
+    case UIAction.sendAllQueuedMessages:
+      for (var conversation in queuedMessagesPerConversation.keys) {
+        for (var message in conversation.messages.where((m) => m.status == model.MessageStatus.inOutbox)) {
+          sendReply(message, conversation);
+        }
+      }
```
danrubel, line 691: `queuedMessagesPerConversation.clear();` before `break;` ?

webapp/lib/controller.dart
```diff
@@ -1002,27 +1159,61 @@ void sendReply(model.SuggestedReply reply, model.Conversation conversation) {
...
       conversation.messages.indexOf(newMessage),
       translation: newMessage.translation,
+      status: newMessage.status,
       incoming: false);
   view.conversationPanelView.addMessage(newMessageView);
+
+  if (!queuedMessagesPerConversation.containsKey(conversation)) {
+    queuedMessagesPerConversation[conversation] = [];
+  }
+  queuedMessagesPerConversation[conversation].add(newMessage);
```
danrubel, line 1169: If you prefer, these lines can be changed to
```
queuedMessagesPerConversation.putIfAbsent(conversation, () => []).add(newMessage)
```

marianamarasoiu: nice, thanks for the suggestion, done!

webapp/lib/controller.dart
marianamarasoiu: I was thinking it would be good to keep the messages in the outbox until they've been confirmed to have been sent successfully, and only then remove them. So I'd rather not clear the list here, but rely on `sendReply` to remove items from this list as it's receiving failure/confirmation notifications.

It's not so clear from the name of this map that it actually holds the conversations in the outbox, not just queued messages for sending (I changed the approach half way through). I'll rename it to make it a bit clearer what its role is, so thanks for highlighting it!

danrubel: @marianamarasoiu Ah! That make sense. Thanks for the explanation.

PR #345, opened 2020-04-24
Add MIT License

ktkization:


ktkization:
@marianamarasoiu @lukechurch SInce I can't add you as reviewers in Nook here's is the PL

PR #346, opened 2020-04-26
Use utc when creating new needs_reply_metrics data

marianamarasoiu:


tool/compute_tag_analytics.py
```diff
@@ -240,7 +240,7 @@ def get_last_incoming_message_dt(messages):
         needs_reply_messages_by_date[day_date] += 1
 
     needs_reply_metrics = {
-        "datetime": datetime.now().isoformat(),
+        "datetime": datetime.now(timezone.utc).isoformat(),
```
danrubel, line 243: Are these two the same and which is preferred?
* `datetime.now(timezone.utc)`
* `datetime.utcnow()`

If they are indeed the same, it would be great if we could standardize on one of them.

marianamarasoiu: they provide the same date, but the difference is between python's timezone aware/naive datetimes. If a timezone is not passed on creation, then it will create a datetime in either local or utc, depending if now() or utcnow() is being used, but it will later not know which timezone it is in. And because of this it can't later compare between naive and aware datetimes.

In other words:
```
>>> datetime.now().isoformat()
'2020-05-20T22:48:11.295866'
>>> datetime.utcnow().isoformat()
'2020-05-20T21:48:13.596408'
>>> datetime.now(timezone.utc).isoformat()
'2020-05-20T21:48:16.020566+00:00'
```

I suggest we go with the timezone aware UTC dates, ie. `datetime.now(timezone.utc)`? Mainly for the extra indication in string form that the date is in UTC.

PR #350, opened 2020-05-03
Update cloud functions

lukechurch:
This brings in the code that was reviewed elsewhere and updates the deployment function to match.


tool/deploy_webapp.sh
```diff
@@ -104,12 +95,30 @@ echo "{\"metadata\": [ {$DEPLOY_DATA} ] }" > public_metadata_nook_app.json
 
 # Deploy using the local firebase tool
```
marianamarasoiu, line 96: Can you also update this comment as it's now obsolete? Thanks!
```suggestion
# Deploy using the firebase tool
```

PR #351, opened 2020-05-05
Firestore helper methods

ktkization:


ktkization:
@marianamarasoiu this is ready for code review and QA

tool/firebase_util.py
```diff
@@ -1,14 +1,63 @@
...
+        field_data = {}
+        for field in data:
+            if field in sub_collections:
+                # it's not actually a field, it's a collection, process these later
+                continue
+            field_data[field] = data[field]
+        firebase_client.document(ref_path).set(field_data)
+
+        for sub_collection in sub_collections:
+            push_collection_to_firestore(f"{ref_path}/{sub_collection}", data[sub_collection])
```
marianamarasoiu, line 55: can you extract this in a separate method like you did for `_push_document_fields_to_firestore`? that way we have a clearer separation between the two types of data, and `push_document_to_firestore` just makes the choice between them.

tool/firebase_util.py
```diff
@@ -1,14 +1,63 @@
...
+    else:
+        _push_document_fields_to_firestore(collection_root, data)
+
+def _push_document_fields_to_firestore(collection_root, data):
```
marianamarasoiu, line 59: can you add a comment for this function and the new `_push_document_to_firestore` function with a sample snippet of how the data looks like for each of them? I think it would help explain what the difference between them is.

tool/firebase_util.py
ktkization: Extracted it to separate method

tool/firebase_util.py
ktkization: Added doc comments

marianamarasoiu: great, thanks!

tool/firebase_util.py
```diff
@@ -1,14 +1,86 @@
...
+      "field1": "",
+      "field2": "",
+    }
+  }
+"""
+def _push_document_fields_to_firestore(collection_root, data):
+    doc_id = list(data.keys())[0]
+    doc_fields = data[doc_id]
+    ref_path = f"{collection_root}/{doc_id}"
+    firebase_client.document(ref_path).set(doc_fields)
```
danrubel, line 86: @ktkization @marianamarasoiu 
I read the `push_document_fields_to_firestore` method name and jumped to the incorrect conclusion that this was updating individual fields and not setting the entire document. I see now that I was wrong. You are just trying to support two different data structures to do the same thing. Please ignore my earlier comments. LGTM and sorry for the noise.

marianamarasoiu:
Hi @danrubel thanks for your earlier detailed review and sorry for the confusion about the two methods - yes, the goal was to support two different data structures.

However, I think your point about the error proneness of the current strategy of choosing which push method to use based on the data structure being supplied is still relevant, even if the two methods aim to do the same thing. So here is a proposal for how to fix that:

We keep `push_collection_to_firestore` and `push_document_to_firestore` (renamed from  `_push_document_to_firestore`), and write a new method (called something like `convert_to_firestore_format`) that takes the data format that `_push_document_fields_to_firestore` takes and transforms it into what `push_document_to_firestore` expects. This way we have a single path for uploading a collection to firebase, that expects data in a specific format (the backup format that's used by `firebase_to_json.py` and `json_to_firebase.py`), and the work of translating other data structures into that format is done separately and beforehand (in this case, for the manually created json that is being used by `add_to_firebase.py` to add suggested replies out of a spreadsheet).

Any thoughts/feedback on this approach? Thanks!

ktkization:
@danrubel  After discussion with @marianamarasoiu I've implemented it as it has been proposed above. Please review the new iteration and give feedback on whether it resolves the problems you've stated in the above comments

tool/firebase_util.py
```diff
@@ -1,14 +1,67 @@
...
+    firestore_format = {}
+    if "__id" in data:
+        firestore_format = data
+    else:
+        doc_id = list(data.keys())[0]
+        firestore_format["__id"] = doc_id
+        firestore_format["__reference_path"] = f"{collection_root}/{doc_id}"
+        firestore_format["__subcollections"] = []
+        firestore_format.update(data[doc_id])
+    return firestore_format
```
marianamarasoiu, line 67: I don't think this is exactly what I suggested. The idea is to remove the `if "__id" in data` check because it's making a decision that should be made earlier in the process. So this method should be called by `add_to_firebase,py` before calling `push_collection_to_firestore`, and not by `push_collection_to_firestore` itself.

```suggestion
def convert_documents_to_firestore_format(collection_root, documents):
    firestore_documents = []
    for doc in documents:
        firestore_doc = {}
        doc_id = list(doc.keys())[0]
        firestore_doc["__id"] = doc_id
        firestore_doc["__reference_path"] = f"{collection_root}/{doc_id}"
        firestore_doc["__subcollections"] = []
        firestore_doc.update(doc[doc_id])
        firestore_documents.append(firestore_doc)
    return firestore_documents
```

[I haven't tested this code, so please test that it's valid python]

tool/firebase_util.py
```diff
@@ -1,14 +1,67 @@
...
     log.info("Done")
     return firebase_client
+
+def push_collection_to_firestore(collection_root, documents):
+    log.info (f"push_collection_to_firestore {collection_root}")
+    col = firebase_client.collection(collection_root)
+    time_start = time.perf_counter_ns()
+    for document_dict in documents:
+        firestore_format_document_dict = _convert_to_firestore_format(collection_root, document_dict)
+        push_document_to_firestore(collection_root, firestore_format_document_dict)
```
marianamarasoiu, line 30: The conversion should happen before the method is called - `push_collection_to_firestore` should get only a single document format

```suggestion
        push_document_to_firestore(collection_root, firestore_format_document_dict)
```

tool/add_to_firebase.py
```diff
@@ -85,7 +69,7 @@ def text_to_doc_id(prefix, text):
         
         short_id = tool_utils.short_id()
         log.audit(f"add_to_firebase, suggested_replies: JobID ({short_id}), keys to download {json.dumps(suggested_replies_documents)}")
-        push_collection_to_firestore(suggested_replies_collection, suggested_replies_documents)
+        firebase_util.push_collection_to_firestore(suggested_replies_collection, suggested_replies_documents)
```
marianamarasoiu, line 72:
```suggestion
        suggested_replies_documents = firebase_util.convert_documents_to_firestore_format(suggested_replies_collection, suggested_replies_documents)
        firebase_util.push_collection_to_firestore(suggested_replies_collection, suggested_replies_documents)
```

tool/add_to_firebase.py
```diff
@@ -120,7 +104,7 @@ def text_to_doc_id(prefix, text):
 
         short_id = tool_utils.short_id()
         log.audit(f"add_to_firebase, conversation_tags: JobID ({short_id}), keys to download {json.dumps(tags_documents)}")
-        push_collection_to_firestore(tags_collection, tags_documents)
+        firebase_util.push_collection_to_firestore(tags_collection, tags_documents)
```
marianamarasoiu, line 107:
```suggestion
        tags_documents = firebase_util.convert_documents_to_firestore_format(tags_collection, tags_documents)
        firebase_util.push_collection_to_firestore(tags_collection, tags_documents)
```

ktkization:
@marianamarasoiu I had misunderstood your approach. Thanks for the suggested refactoring. I have tested the code and fixed a few bugs I've found. It can be reviewed again.

PR #352, opened 2020-05-06
Script to monitor size of directory

ktkization:


ktkization:
@marianamarasoiu Ready for review and QA

tool/monitor_data_dir.py
```diff
@@ -0,0 +1,33 @@
...
+    parser.add_argument("crypto_token_file", type=str, help="path to Firebase crypto token file")
+    parser.add_argument("project", type=str, help="name of the project")
+    parser.add_argument("dirpath", type=str, help="path of directory to monitor")
+    args = parser.parse_args()
+    
+    CRYPTO_TOKEN_PATH = args.crypto_token_file
+    dirpath = os.path.abspath(args.dirpath)
+    project = args.project
+    firebase_client = init_firebase_client(CRYPTO_TOKEN_PATH)
+    runner = threading.Thread(target=run)
```
marianamarasoiu, line 32: I don't think we need a new thread for this, let's just run it directly?

tool/data_dir_stat.py
```diff
@@ -0,0 +1,30 @@
...
+log = None
+
+def get_size_and_upload(CRYPTO_TOKEN_PATH, dirpath, project, firebase_client):
+    global log
+    if log is None:
+        log = logging.Logger(__file__, CRYPTO_TOKEN_PATH)
+    dirname = os.path.basename(dirpath)
+    dir_size = get_dir_size(dirpath)
+    entry = {
+        'project': project,
```
marianamarasoiu, line 15: can you also add the dirname in the entry as well?

tool/data_dir_stat.py
```diff
@@ -0,0 +1,30 @@
...
+        'size': round(dir_size, 1)
+    }
+    collection = dirname + '_dir_size_metrics'
+    timestamp = datetime.datetime.now().isoformat()
+    firebase_client.collection(collection).document(timestamp).set(entry)
+    log.info('Pushed size of directory {} successfully to Firestore'.format(dirname))
+
+# size in MB
+def get_dir_size(dirpath):
```
marianamarasoiu, line 24: consider adding "in_mb" to the function name, that way you won't need the comment

tool/data_dir_stat.py
```diff
@@ -0,0 +1,30 @@
...
+
+def get_size_and_upload(CRYPTO_TOKEN_PATH, dirpath, project, firebase_client):
+    global log
+    if log is None:
+        log = logging.Logger(__file__, CRYPTO_TOKEN_PATH)
+    dirname = os.path.basename(dirpath)
+    dir_size = get_dir_size(dirpath)
+    entry = {
+        'project': project,
+        'size': round(dir_size, 1)
```
marianamarasoiu, line 16: consider using "size_in_mb" for clarity here as well

tool/data_dir_stat.py
```diff
@@ -0,0 +1,30 @@
...
+    }
+    collection = dirname + '_dir_size_metrics'
```
marianamarasoiu, line 18: consider keeping the collection name as "dir_size_metrics" and adding the dirname as part of the entry

tool/monitor_data_dir.py
ktkization: Ok. I have refactored it out

tool/data_dir_stat.py
ktkization: Added it

tool/data_dir_stat.py
ktkization: Refactored it to that

tool/data_dir_stat.py
ktkization: Refactored it to that

tool/data_dir_stat.py
ktkization: Renamed it to that

ktkization:
Moved the script into a single file

tool/monitor_directory.py
```diff
@@ -0,0 +1,52 @@
...
+from katikati_pylib.logging import logging
+
+CRYPTO_TOKEN_PATH = None
+DEFAULT_INTERVAL = 600 # 10 minutes
+firebase_client = None
+log = None
+
+def get_size_and_upload(dirpath, project):
+    global log
+    if log is None:
```
marianamarasoiu, line 16: I think it makes more sense now to create the logger under main rather than here since this will be run as a standalone script rather than being imported somewhere else. There will also be no need then to keep a global reference to `CRYPTO_TOKEN_PATH`

tool/monitor_directory.py
```diff
@@ -0,0 +1,53 @@
...
+    parser = argparse.ArgumentParser(description="Monitor size of a directory")
+    parser.add_argument("crypto_token_file", type=str, help="path to Firebase crypto token file")
+    parser.add_argument("project", type=str, help="name of the project")
+    parser.add_argument("dirpath", type=str, help="path of directory to monitor")
+    args = parser.parse_args()
+    
+    dirpath = os.path.abspath(args.dirpath)
+    project = args.project
+    
+    global log
```
marianamarasoiu, line 45: you can't have `global log` here - it's not in a function, but in the top-level namespace. I think python will actually crash if we try to run this. You also don't need to check if it's `None`, it will be

So just this will be enough:
`log = logging.Logger(__file__, args.crypto_token_file)`

PR #353, opened 2020-05-07
Display latest git commit sha

ktkization:


ktkization:
@marianamarasoiu This is ready for Review & QA

webapp/lib/platform_constants.dart
```diff
@@ -4,13 +4,17 @@ import 'dart:convert';
 import 'dart:html';
 
 String _constantsFilePath = 'assets/firebase_constants.json';
+String _latestCommitHashConfigFilePath = 'assets/latest_commit_hash.json';
```
marianamarasoiu, line 7: I don't think this should be in the platform part of the codebase, as platform is used here to refer to the firebase and cloud pubsub services. So I think it fits better for this file to be read directly in the view part of the app, and also without the controller intermediation as there's no user interaction about it.

ktkization: Ok. I've moved the logic into the view

webapp/lib/view.dart
```diff
@@ -44,16 +45,19 @@ void init() {
   document.onKeyDown.listen((event) => command(UIAction.keyPressed, new KeyPressData(event.key)));
 }
 
-void initSignedInView() {
+void initSignedInView() async {
```
marianamarasoiu, line 48: We need to be a bit careful with turning methods into an async ones, and I'd rather us not do that here. If for some reason Nook can't read the commit hash file, it will crash, and the UI won't load at all. Reading the hash is more for debugging than core functionality - Nook should be able to run without it.

Can you instead leave the signed in status as it was, without the hash, and afterwards attempt to read the file and replace `showNormalStatus` with the commit hash?

ktkization: Refactored this to get hash afterwards

PR #355, opened 2020-05-12
Add timestamp field for charting purpose

ktkization:


ktkization:
@marianamarasoiu Please review this when you have time

PR #359, opened 2020-05-19
Make it harder to send multi messages with keyboard shortcuts

marianamarasoiu:
This PR does two things:
- it disables multi-message send with shortcuts - the user now has to use the explicit send button
- if a modifier key is pressed at the same time as another key, we don't send the matching suggested reply - this is to avoid sending messages when the user wants to use typical keyboard shortcuts in chrome (ctrl-c/ctrl-v/ctrl-0 etc)

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -182,7 +182,8 @@ class UserData extends Data {
 
 class KeyPressData extends Data {
   String key;
-  KeyPressData(this.key);
+  bool hasModifierKey;
+  KeyPressData(this.key, this.hasModifierKey);
 
   @override
   String toString() => 'KeyPressData: {key: $key}';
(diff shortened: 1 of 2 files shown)


webapp/lib/controller.dart
```diff
@@ -821,17 +822,14 @@ void command(UIAction action, Data data) {
         view.snackbarView.hideSnackbar();
       }
       // If the shortcut is for a reply, find it and send it
-      var selectedReply = suggestedRepliesByCategory[selectedSuggestedRepliesCategory].where((reply) => reply.shortcut == keyPressData.key);
+      var selectedReply = suggestedRepliesByCategory[selectedSuggestedRepliesCategory].where((reply) => reply.shortcut == keyPressData.key && !keyPressData.hasModifierKey);
```
danrubel, line 825: Rather than this change, perhaps wrap lines 825 - 834 in
```
if (!keyPressData.hasModifierKey) {
    ...
}
```
so that the "hasModifierKey prevents all replies" is more obvious?

marianamarasoiu: Thanks @danrubel that's indeed more straightforward. On second thought, we also want to ignore modifier key-shortcuts for tags as well, so I replaced this with an early exit if a modifier key is present. PTAL

danrubel: @marianamarasoiu Good change. I like that even better.

PR #360, opened 2020-05-19
Add a filterable field to tags to reduce the filter list

marianamarasoiu:
And also consolidate the filtering in the _populateTagView methods.

Temporarily consider the field optional - will make it required and remove the null check once the infrastructure is updated to push the field to firebase.

### DIFF
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -349,6 +349,7 @@ class Tag {
   String text;
   TagType type;
   String shortcut;
+  bool filterable;
 
   String get tagId => docId;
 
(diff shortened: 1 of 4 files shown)


webapp/lib/model.g.dart
```diff
@@ -349,6 +349,7 @@ class Tag {
   String text;
   TagType type;
   String shortcut;
+  bool filterable;
```
danrubel, line 352: Wouldn't hurt to add this to `model.yaml` as well... but not required

marianamarasoiu: Good point, done, thanks!

Issue #349, opened 2020-04-29
Tag filtering UI is unwieldy again


lukechurch:
We need to rethink through the tag filtering UI again - removing the demogs tags got us some way but the complexity of the projects has resulted in this becoming over-full again

marianamarasoiu:
Closing as resolved

PR #372, opened 2020-05-27
Add implementation for filtering hierarchical tags

marianamarasoiu:
This PR addresses #349 by:
- adding categories for tags in the same way that we have added for suggested replies
- using the categories in the filtering UI via a category selector
- also setting an upper limit on the height of the tag filtering menu making the list to scroll if there are too many tags, this should make it more usable even when we're not using categories

### DIFF
--- a/webapp/lib/controller_view_helper.dart
+++ b/webapp/lib/controller_view_helper.dart
@@ -156,7 +156,7 @@ Map<String, List<model.SuggestedReply>> _groupRepliesIntoCategories(List<model.S
 Map<String, List<model.Tag>> _groupTagsIntoCategories(List<model.Tag> tags) {
   Map<String, List<model.Tag>> result = {};
   for (model.Tag tag in tags) {
-    result.putIfAbsent(tag.category, () => []).add(tag);
+    result.putIfAbsent(tag.group, () => []).add(tag);
   }
   return result;
 }
--- a/webapp/lib/controller_view_helper.dart
+++ b/webapp/lib/controller_view_helper.dart
@@ -152,3 +152,15 @@ Map<String, List<model.SuggestedReply>> _groupRepliesIntoCategories(List<model.S
   }
   return result;
 }
+
+Map<String, List<model.Tag>> _groupTagsIntoCategories(List<model.Tag> tags) {
+  Map<String, List<model.Tag>> result = {};
+  for (model.Tag tag in tags) {
+    String category = tag.category ?? '';
+    if (!result.containsKey(category)) {
+      result[category] = [];
+    }
+    result[category].add(tag);
+  }
+  return result;
+}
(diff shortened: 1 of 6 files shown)


webapp/lib/controller_view_helper.dart
```diff
@@ -152,3 +152,15 @@ Map<String, List<model.SuggestedReply>> _groupRepliesIntoCategories(List<model.S
   }
   return result;
 }
+
+Map<String, List<model.Tag>> _groupTagsIntoCategories(List<model.Tag> tags) {
+  Map<String, List<model.Tag>> result = {};
+  for (model.Tag tag in tags) {
+    String category = tag.category ?? '';
+    if (!result.containsKey(category)) {
```
danrubel, line 160: Another option... `result.putIfAbsent(category, () => []).add(tag);`
Only involves 1 hash map lookup rather than 2.

webapp/lib/model.g.dart
```diff
@@ -371,7 +372,8 @@ class Tag {
       ..text = String_fromData(data['text'])
       ..type = TagType.fromString(data['type'] as String) ?? TagType.Normal
       ..shortcut = String_fromData(data['shortcut'])
-      ..filterable = bool_fromData(data['filterable']);
+      ..filterable = bool_fromData(data['filterable'])
+      ..category = String_fromData(data['category']);
```
danrubel, line 376: Does it make sense for `category` to always have a non-null value?
If so... you could append `?? ''` to this line and remove null checks elsewhere.

webapp/lib/controller_view_helper.dart
marianamarasoiu: So much better, I don't know why I keep forgetting that `putIfAbsent` exists! Thanks @danrubel!

I'll update the similar code for suggested replies as well in a future PR

webapp/lib/model.g.dart
marianamarasoiu: Yes, I think it does. And might make sense to have done the same for `filterable`. I'll follow up with another PR for this. thanks @danrubel !

Speaking of - at the moment I don't think the UI has any checks as to what is optional in the data structure (and so can have a default value) and what is required. I believe that we're primarily relying on the backend infrastructure to ensure the data is in the right format. Would it make sense to add some of those checks on the UI side as well? @danrubel @lukechurch what do you think?

danrubel: @marianamarasoiu Good question. We can easily add default values to replace nulls or missing data fields and there is some backend to UI data coercion in place (e.g. [int_fromData](https://github.com/larksystems/nook/blob/d1c902201d85023bd6c5c6f6c3911e850a7fa84c/webapp/lib/model.g.dart#L536)), but you are right that we don't have any UI checks on data sent to the backend other than Dart types themselves.

@lukechurch Thoughts?

lukechurch: I think it makes sense for us to add verification checks on the importing of the data - mainly to improve the error reporting and localisation. I'm not sure that we want the UI trying to do much repair, but checking makes sense.

webapp/lib/model.yaml
```diff
@@ -51,6 +51,7 @@ Tag:
   type: 'TagType, TagType.Normal'
   shortcut: 'string'
   filterable: 'bool'
+  category: 'string'
```
lukechurch, line 54: per discussion let's call this 'group'

marianamarasoiu: thanks, done!

PR #374, opened 2020-05-28
SuggestedReply.setTranslation via add_opinion

danrubel:
This updates `SuggestedReply.setTranslation` to use `add_opinion`.

If this change looks correct, please merge when appropriate piety changes have been made.


lukechurch:
As per the comment on the other PR - we've changed the way we're handling suggestedReplies in Nook - so these should be read only.

danrubel:
Good to know. Closing without merging.

PR #375, opened 2020-05-29
remove unused functionality

danrubel:
Per comments in https://github.com/larksystems/nook/pull/374, this functionality is not used now and is not intended to be used in the future.


webapp/lib/model.yaml
```diff
@@ -51,7 +51,7 @@ Tag:
   type: 'TagType, TagType.Normal'
   shortcut: 'string'
   filterable: 'bool'
-  group: 'string'
+  group: "string, ''"
```
marianamarasoiu, line 54: thanks for this @danrubel !

PR #380, opened 2020-06-03
Various layout fixes

### LINKED ISSUE #379: Long messages in the conversation panel still causes the UI to reflow

marianamarasoiu:
* Fixes #379 which was caused by the not removing the width classes between tag panel visibility changes
* Long tag names without spaces in the tag panel will break on multiple lines if there's not enough space for them in the panel
* "Tag" buttons are also now floating over the text of the tag (as for the suggested replies)

### DIFF
--- a/webapp/lib/view.dart
+++ b/webapp/lib/view.dart
@@ -84,18 +84,35 @@ void clearMain() {
 }
 
 void showTagPanel() {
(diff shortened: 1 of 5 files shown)


lukechurch:
Great to have these

marianamarasoiu:
thanks both! merging..

PR #381, opened 2020-06-07
Rewrite the group tags UI to show all tags in foldable menus

marianamarasoiu:
This removes the need to keep track in the controller of which group is being shown currently.

Re #349

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -50,7 +50,6 @@ enum UIAction {
   deselectAllConversations,
   updateSystemMessages,
   updateSuggestedRepliesCategory,
-  updateFilterTagsCategory,
   hideAgeTags,
   showSnackbar
 }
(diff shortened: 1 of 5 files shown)


webapp/lib/controller.dart
```diff
@@ -310,32 +307,16 @@ void initUI() {
...
-      view.conversationFilter.categories = categories;
-      // If the categories have changed under us and the selected one no longer exists,
-      // default to the first category, whichever it is
-      if (!categories.contains(selectedFilterTagsCategory)) {
-        selectedFilterTagsCategory = categories.first;
-      }
-      // Select the selected category in the UI and add the tags for it
-      view.conversationFilter.selectedCategory = selectedFilterTagsCategory;
-      _populateFilterTagsMenu(filterTagsByCategory[selectedFilterTagsCategory]);
+      _removeTagsToFilterMenu(_groupTagsIntoCategories(removed));
```
danrubel, line 310: rename to `_removeTagsFromFilterMenu` ?

marianamarasoiu: thanks @danrubel , done!

PR #386, opened 2020-06-09
Fix errors thrown when the active conversation is no longer in the list view

### LINKED ISSUE #383: Conversation update doesn't update the conversation tags
Repro:

- Find a conversation
- Add a message tag that results in the engine updating the conversation tags
- The conversation tags remain the same, even though e.g. the filter warning works correctly

marianamarasoiu:
Fixes #383 

By keeping the active conversation on the screen even when it no longer meets filtering constraints, the `activeConversation.docId` may not be in the `view.conversationListPanelView`. This PR addresses this by checking if it is there before trying to select it and adding null checks for markUnread/check operations.


marianamarasoiu:
thanks @lukechurch 

Merging this into master ahead of deployment to the projects - @danrubel please do have a look even after it's merged and if you catch anything I'll address it in a future PR. thanks!

Issue #377, opened 2020-06-02
Read only configuration for Nook


lukechurch:
We'll need a user configuration for Nook that disables all edits and sends - so a 'viewer' mode. We may also need a mode that can edit tags, but not send messages, unclear at the moment.

Needed for a project going live on 21st June

lukechurch:
On reflection I think we don’t need a read only mode, we need something a bit finer grained.

Can send custom messages?
Can send messages?
Show suggested replies column?
Can edit translations?
Can add remove message tags?
Can add remove conversation tags?
Can create new tags? (!)
Can edit notes? (edited)

marianamarasoiu:
closing as resolved

PR #387, opened 2020-06-11
Add more user config settings

marianamarasoiu:
Addresses the configs listed in #377, except the one for adding new replies and tags, as that would require more implementation work other than toggling it on or off. I have actually commented the respective buttons out, as they didn't really work. Will add them back in when they're ready.

A couple of things to pay more attention to:
- I've been going back and forth on this, but in the end decided that the controller should handle all config settings, including making the decision for the view. This means that the view constructors will start with an easy default choice first (e.g. disable editable translations, but showing reply/tag buttons), then when the controller is initialised, it will explicitly apply the default configs (all settings set to false). And when the configuration is loaded from firebase, then it gets propagated to the view. This way, the view doesn't have to keep checking the configs as to what it should do, and it's the job of the controller to handle the logic. This makes the view simpler, but thinking about the controller code will be a bit more complicated in that we need to remember all the ways in which it needs to initialise the view.

- with the replies and tags panels being optional now, there are 4 layout options, and the panel widths all vary between the 4 options. I've basically created a (hard-coded) matrix of widths for this as using something like flexbox wasn't being stable enough for us (e.g. showing long messages without any spaces between words). If you have any other suggestions for how to do this, or a nicer way to define and then select the layout, please let me know.

Thanks!


webapp/lib/model.g.dart
```diff
@@ -250,6 +250,7 @@ class MessageStatus {
         if (value.name == valueName) return value;
       }
     }
+    log.warning('unknown MessageStatus $text');
```
danrubel, line 253: This is generated, but Luke indicated that we should remove this because it was generating too much noise in the logs. Please comment this out and add a comment above it along the lines of 
`// This is commented out because it generates too much noise in the logs`

I've just been manually removing this line each time I regenerate the model.
I should have commented out the line and added an explanation. Sorry about that.

webapp/lib/controller.dart
```diff
@@ -422,15 +422,24 @@ void initUI() {
...
+    newConfig.repliesKeyboardShortcutsEnabled ? view.replyPanelView.showShortcuts() : view.replyPanelView.hideShortcuts();
+  }
+
+  if (currentConfig.tagsKeyboardShortcutsEnabled != newConfig.tagsKeyboardShortcutsEnabled) {
+    newConfig.tagsKeyboardShortcutsEnabled ? view.tagPanelView.showShortcuts() : view.tagPanelView.hideShortcuts();
+  }
+
+  if (currentConfig.sendMessagesEnabled != newConfig.sendMessagesEnabled) {
+    newConfig.sendMessagesEnabled ? view.replyPanelView.showButtons() : view.replyPanelView.hideButtons();
   }
```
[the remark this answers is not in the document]
marianamarasoiu: The view doesn't remember its show/hide state, but even so, I like your API design much better, I think it makes the code more straightforward, so I've now updated all of the methods to use it instead of the show/hide style. Thanks!

webapp/lib/model.g.dart
marianamarasoiu: Ah, yes, I think I remember seeing a discussion along these lines, I should have dug it up before just adding the line in. Commented out now with a comment as you suggested. thanks!

webapp/lib/controller.dart
```diff
@@ -747,6 +781,7 @@ void command(UIAction action, Data data) {
       activeConversation = null;
       view.conversationListPanelView.clearConversationList();
       view.conversationPanelView.clear();
+      view.replyPanelView.noteText = '';
```
lukechurch, line 784: Can I just check this change was intended? We don't use this field much so far, so would like to make sure that it's not a stray bug that's going to overwrite the field?

marianamarasoiu: Yes, this was intended, and it won't overwrite the field, it will only clear the text of the notes HTML element.

What happened was that the notes field wasn't being reset when switching between conversation lists. Most of the times this isn't an issue, as the notes view gets set when a new conversation is selected, and that's by default the first one in the list after a switch. The only time this would be an issue is when switching from one of the lists to the default of no list selected (the "Select the conversation list to be displayed" in the dropdown). 

Hope this makes sense? And sorry, I came across this small bug during testing and so I just fixed it in place - it would have been clearer to do this in a separate PR.

lukechurch:
Yup, that makes sense - and no worries about cleaning up small stuff like
this as you see it

PR #388, opened 2020-06-14
Explicitly configure OAuth domains to use and add University of Cambridge login

marianamarasoiu:


lukechurch:
Thanks @marianamarasoiu, LGTM, following up on internal comms

webapp/lib/controller.dart
```diff
@@ -805,7 +820,8 @@ void command(UIAction action, Data data) {
       initUI();
       break;
     case UIAction.signInButtonClicked:
-      platform.signIn();
+    SignInData signInData = data;
```
danrubel, line 823: Indent?

marianamarasoiu: thanks for the catch @danrubel, fixed!

PR #391, opened 2020-06-27
Firebase rules to allow University of Cambridge email domains

marianamarasoiu:


lukechurch:
LGTM, but I would like this change to only be deployed to nook-1 whilst we do careful testing, thanks

marianamarasoiu:
Thanks both. I'll close this PR rather than merge, as per discussion with @lukechurch , and will deploy this change separately to nook-1.

PR #395, opened 2020-07-09
Don't try to remove tags that don't exist.

marianamarasoiu:
This shouldn't happen, but the missing tag lists don't seem to work well, causing attempts to remove tags that are not the the view. It's a stop gap fix until we find the proper cause for this issue.


marianamarasoiu:
Deploying to one of the projects that is encountering this issue, but not merging as I'd prefer to have the proper fix in instead of this.

marianamarasoiu:
Closing without merging - this is properly fixed in #398

PR #398, opened 2020-07-09
Fix issue where tags were removed from view but not app controller state

marianamarasoiu:
It turns out the proper fix for #395 is fairly simple - we were only removing them from the view, but not from the internal app state.


lukechurch:
Thanks @marianamarasoiu!

PR #400, opened 2020-07-13
Group tags in RHS panel and show groups in dropdown selector

### LINKED ISSUE #385: Support tags appearing in multiple groups
At the moment tags can only appear in one group - but they're used between different seasons - so it would help to be able to express a list of groups for a tag

marianamarasoiu:
The UI (and implementation approach) is very similar to what we do for suggested replies.

Closes #385


lukechurch:
LGTM with minor nits as discussed

PR #402, opened 2020-07-21
Script to migrate nook watcher data

ktkization:


tool/migrate_nook_watcher_data.py
```diff
@@ -0,0 +1,53 @@
...
+
+firebase_client = None
+log = None
+
+def strip_project(data):
+    for entry in data:
+        try:
+            del list(entry.values())[0]['project']
+        except Exception:
+            pass
```
marianamarasoiu, line 15: I'm always skeptical of catching exceptions and doing nothing with them...
What kind of exceptions are we likely to get here? that the project field doesn't exist? are they likely to be one-offs, or for entire collections? would it be worth logging it as a warning at least?

tool/migrate_nook_watcher_data.py
```diff
@@ -0,0 +1,53 @@
...
+    return data
+
+def migrate():
+    if args.project:
+        docs = firebase_client.collection(args.source).where('project', '==', args.project).stream()
+    else:
+        docs = firebase_client.collection(args.source).stream()
+    data = [{doc.id: doc.to_dict()} for doc in docs][-20:]
```
marianamarasoiu, line 23: This seems to be selecting only the last 20 docs, which doesn't seem right to me? is it an artefact from testing or am I misunderstanding something?

tool/migrate_nook_watcher_data.py
```diff
@@ -0,0 +1,53 @@
...
+        firebase_client.collection(collection_path).document(doc_label).set(doc_data)
+    log.info('done migration')
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawTextHelpFormatter,
+        epilog=textwrap.dedent("""
+        Examples:
+        python migrate_nook_watcher_data.py [CRYPRO_TOKEN_PATH] needs_reply_metrics Lark_KK-Project-2020-COVID19-KE-URBAN/needs_reply --project Lark/KK-Project-2020-COVID19
```
marianamarasoiu, line 39: for consistency:
```suggestion
        python migrate_nook_watcher_data.py [CRYPRO_TOKEN_PATH] needs_reply_metrics Lark_KK-Project-2020-COVID19-KE-URBAN/needs_reply --project Lark/KK-Project-2020-COVID19-KE-URBAN
```

tool/migrate_nook_watcher_data.py
```diff
@@ -0,0 +1,53 @@
...
+        python migrate_nook_watcher_data.py [CRYPRO_TOKEN_PATH] system_events Lark_KK-Project-2020-COVID19-KE-URBAN/system_events --project Lark/KK-Project-2020-COVID19
```
marianamarasoiu, line 40: as above 
```suggestion
        python migrate_nook_watcher_data.py [CRYPRO_TOKEN_PATH] system_events Lark_KK-Project-2020-COVID19-KE-URBAN/system_events --project Lark/KK-Project-2020-COVID19-KE-URBAN
```

tool/migrate_nook_watcher_data.py
ktkization: It's from testing. Removed it

tool/migrate_nook_watcher_data.py
ktkization: Since there some earlier documents without the project property, deleting a non-existent property in a dictionary throws an exception. As per your suggestion, I've decided to define the specific exception and show a warning if it doesn't exist

PR #405, opened 2020-07-28
Keep selected conversations in view with a warning when they no longer meet filters

marianamarasoiu:


webapp/lib/view.dart
```diff
@@ -1137,6 +1147,10 @@ class ConversationSummary with LazyListViewItem {
     _checkboxHidden = !show;
     if (_selectCheckbox != null) _selectCheckbox.hidden = !show;
   }
+  void _showWarning(bool show) {
+    _warning = true;
```
danrubel, line 1151: `_warning = show;` ?

marianamarasoiu: Yes, sorry, fixed now, thanks for catching this, it would have been quite a pain to debug!

lukechurch:
Thanks @marianamarasoiu - can we bring this in to start testing?

lukechurch:
@marianamarasoiu - sorry this seems to have drifted, could you bring back up to head please. Ideally I'd like to start testing this over the next few days

PR #411, opened 2020-08-04
Add exclude filter and reuse shared code between filters

marianamarasoiu:
This is a big and messy change, sorry in advance...

Basically it brings back the exclude filter prototyped a while back ( #194 and #208). With quite a bit of the code needing to be shared between the (now) three filters, I took the earlier approach of grouping the filters in a class and differentiating between them via a `TagFilterType` enum (hence some of the more recent last inbound turn code being deleted).

The CSS changes are to make the management of the stacked filter menus better, by changing the hover border from a rectangle taking the whole left side of the menu to a triangle:
Screenshot 2020-08-04 at 01 08 03

Re #189 (if this change feels ok to use, it might even close that issue!)

Thanks!

### DIFF
--- a/webapp/lib/controller_filter_helper.dart
+++ b/webapp/lib/controller_filter_helper.dart
@@ -1,0 +1,80 @@
+part of controller;
+
+enum TagFilterType {
+  include,
+  exclude,
+  lastInboundTurn
+}
+
+class ConversationFilter {
+  Map<TagFilterType, List<model.Tag>> filterTags;
+  Map<TagFilterType, DateTime> afterDateFilter;
+
+  ConversationFilter() {
+    filterTags = {
+      TagFilterType.include: [],
+      TagFilterType.exclude: [],
+      TagFilterType.lastInboundTurn: []
+    };
+    afterDateFilter = {
+      TagFilterType.include: null,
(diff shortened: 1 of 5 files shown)


webapp/lib/controller_filter_helper.dart
```diff
@@ -0,0 +1,80 @@
...
+    filterTags[TagFilterType.lastInboundTurn] = tagIdsToTags(filterTagIds, messageTags).toList();
+
+    // after date filter
+    afterDateFilter[TagFilterType.include] = view.urlView.getPageUrlFilterAfterDate(TagFilterType.include);
+    afterDateFilter[TagFilterType.exclude] = view.urlView.getPageUrlFilterAfterDate(TagFilterType.exclude);
+  }
+
+  // bool contains(model.Tag tag) {
+  //   return includeTags.contains(tag) || excludeTags.contains(tag);
+  // }
```
danrubel, line 51: Remove?

lukechurch:
This has proved great in practice, thanks very much @marianamarasoiu

marianamarasoiu:
thanks @danrubel and @lukechurch ! merging and deploying

PR #412, opened 2020-08-04
split enum to/fromString and to/fromData

danrubel:
This PR contains several enum change based on new codegen, but no functional changes:
* new enum `fromData` and `toData` methods
* update model classes to use new enum `fromData` and `toData` methods rather than `fromString` and `toString`
* syntax style cleanup of enum `fromString` method

Two reasons behind these changes:
1) The `toString` method in Dart is typically used for debugging and it is better to have a distinct `toData` method for converting enums into data. The enum `toString` and `toData` methods happen to return the same thing today but may diverge in the future.
2) Updating codegen to call enum `fromData` makes it more consistent with all other generated model classes.

### DIFF
--- a/webapp/lib/model.g.dart
+++ b/webapp/lib/model.g.dart
@@ -173,9 +173,9 @@ class Message {
   static Message fromData(data, [Message modelObj]) {
     if (data == null) return null;
     return (modelObj ?? Message())
-      ..direction = MessageDirection.fromString(data['direction'] as String) ?? MessageDirection.Out
+      ..direction = MessageDirection.fromData(data['direction']) ?? MessageDirection.Out
       ..datetime = DateTime_fromData(data['datetime'])
-      ..status = MessageStatus.fromString(data['status'] as String)
+      ..status = MessageStatus.fromData(data['status'])
       ..tagIds = List_fromData<String>(data['tags'], String_fromData)
       ..text = String_fromData(data['text'])
       ..translation = String_fromData(data['translation'])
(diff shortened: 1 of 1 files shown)


webapp/lib/model.g.dart
```diff
@@ -209,7 +209,7 @@ class MessageDirection {
   static MessageDirection fromString(String text, [MessageDirection defaultValue = MessageDirection.Out]) {
     if (text != null) {
       const prefix = 'MessageDirection.';
-      String valueName = text.startsWith(prefix) ? text.substring(prefix.length) : text;
+      var valueName = text.startsWith(prefix) ? text.substring(prefix.length) : text;
```
marianamarasoiu, line 212: has `valueName` changed type / can it be non-String now? I'm fine either way, just wondering

danrubel: @marianamarasoiu No functional or type change... just a Dart syntax style change. The prevalent Dart style is to use `var` when declaring local variables of the form `var x = something` because Dart can intuit the local variable's type from the right hand side of the assignment. That local variable has that type from that point forward and cannot be assigned something of a different type.

That said...
1) Sometimes I use the variable type on the left hand side to make it more obvious
2) If it is more clear to use the variable type here or elsewhere please let me know and I can change it.

marianamarasoiu: yup, that makes sense, thanks for the explanation! I'll also keep the recommended style in mind for future code, I haven't developed that good a relationship with `var` yet :)

Issue #316, opened 2020-04-03
Need to be able to select the shard / conversation in Nook via the URL


lukechurch:

lukechurch:
Also need to be able navigate directly to a conversation via the URL

PR #417, opened 2020-08-06
Add shard ID to page URL

marianamarasoiu:
First part for being able to select a conversation via URL ( #316 )

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -484,6 +484,8 @@ void initUI() {
         log.warning("Attempting to select shard ${conversationListRoot} that doesn't exist");
         conversationListRoot = ConversationListData.NONE;
       }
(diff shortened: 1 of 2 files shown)


webapp/lib/controller.dart
```diff
@@ -474,6 +474,19 @@ void initUI() {
         ..addAll(added)
         ..addAll(modified);
       view.conversationListSelectView.updateConversationLists(shards);
+
+      // Read any conversation shards from the URL
+      String urlConversationListRoot = view.urlView.getPageUrlConversationList();
```
lukechurch, line 479: Do you think this might be better as a helper function?

marianamarasoiu: I'd rather not - it's a bit of a complicated piece of logic, and this is the only place where it needs to be computed, so I'd prefer to keep it in context, for now at least

PR #423, opened 2020-08-24
Add an optional --only flag to pass to firebase deploy

marianamarasoiu:
This will help make Nook deployments faster by being able to specify that only hosting should be deployed (there's other options, but hosting will be most useful).

It will also make deployment to the SOM-CC project easier, as it has a different set of database access rules that we don't maintain in the repo, which have sometimes been accidentally overwritten on deployments.

Tested both with and without the flag in the latest round of Nook deployments.

### DIFF
--- a/tool/deploy_webapp.sh
+++ b/tool/deploy_webapp.sh
@@ -5,6 +5,11 @@ WORK_DIR="$(pwd)"
 
 ########## parse and validate argument(s)
 
+usage_and_exit() {
+  echo "Usage: $0 [--skip-cloud-functions] [--only hosting|database|firestore] path/to/firebase/deployment/constants.json path/to/crypto/token/file"
+  exit 1
+}
+
 FIREBASE_DEPLOY_ARGS=()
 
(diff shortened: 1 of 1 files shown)


tool/deploy_webapp.sh
```diff
@@ -1,17 +1,34 @@
...
+while [ "$#" -gt 0 ]; do
+  case "$1" in
+    --skip-cloud-functions)
+      SKIP_CLOUD_FUNCTIONS=true
+      shift
+      ;;
+    --only)
+      shift
+      FIREBASE_DEPLOY_ARGS[0]="--only"
+      FIREBASE_DEPLOY_ARGS[1]="$1"
```
danrubel, line 19: Perhaps add `if [ -z "$1" ]; then...` here to validate that there is indeed an argument after the `--only`

marianamarasoiu: good idea, thanks @danrubel !

I've also added a `usage_and_exit` function as well and added it in the other places we were exiting with an error.

PR #426, opened 2020-08-30
Add conversation ID to the URL

### LINKED ISSUE #316: Need to be able to select the shard / conversation in Nook via the URL

marianamarasoiu:
Final step for being able to point to a conversation via the URL. Fixes #316

### DIFF
--- a/webapp/lib/model.dart
+++ b/webapp/lib/model.dart
@@ -30,7 +30,13 @@ extension ConversationListShardUtil on g.ConversationListShard {
 }
 
 extension ConversationUtil on g.Conversation {
-  String get shortDeidentifiedPhoneNumber => docId.split('uuid-')[1].split('-')[0];
+  String get shortDeidentifiedPhoneNumber {
+    try {
+      return docId.split('uuid-')[1].split('-')[0];
+    } catch (e) {
+      return docId;
+    }
+  }
 
   /// Return the most recent inbound message, or `null`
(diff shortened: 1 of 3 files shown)


webapp/lib/model.dart
```diff
@@ -30,7 +30,13 @@ extension ConversationListShardUtil on g.ConversationListShard {
 }
 
 extension ConversationUtil on g.Conversation {
-  String get shortDeidentifiedPhoneNumber => docId.split('uuid-')[1].split('-')[0];
+  String get shortDeidentifiedPhoneNumber {
+    try {
+      return docId.split('uuid-')[1].split('-')[0];
+    } catch (e) {
```
lukechurch, line 36: Why would this circumstance be triggered?

lukechurch (approved):

webapp/lib/model.dart
marianamarasoiu: the nook id is now visible in the URL, and so if the url is edited and it no longer meets the "nook-phone-uuid-XXX" format, this will crash. I didn't think this was a very great solution either, so perhaps I should revert this and either catch the error another way, or say, this should never happen - so letting the error through is the right thing to do?

marianamarasoiu:
merging this now so we can do further work on top of it - @lukechurch let's follow up on the discussion about the `shortDeidentifiedPhoneNumber` separately

PR #427, opened 2020-08-30
Add a filter component to search by conversation ID

marianamarasoiu:
Screenshots:

Screenshot 2020-08-30 at 17 20 47
Screenshot 2020-08-30 at 17 21 12

It works with both short and long IDs.

Note: this PR was developed on top of #426 , it will need to be merged after that one, and the target branch changed to master.

PR #428, opened 2020-08-30
Various tag filtering improvements

marianamarasoiu:
* Selected filter tags will be loaded from the URL on the first load of Nook, even with no conversation was selected. They will temporarily appear as their tag ID in yellow, and then will be replaced by their name once the tags have been loaded.
* Fixes handling of after date filters - they are now written and read from the URL properly
* Changes the conversation list panel to say "X/N conversations" - useful when filtering to know the total number of conversations
* Doesn't call `updateFilteredAndSelectedConversationLists` when no conversation list is selected, and filter tags and being adjusted (this was causing the "Select list" message to be cleared)

Note: this PR was developed on top of #427 , it will need to be merged after that one, and the target branch changed to master.

### DIFF
--- a/webapp/lib/view.dart
+++ b/webapp/lib/view.dart
@@ -815,8 +815,8 @@ class ConversationListPanelView {
   Map<String, ConversationSummary> _phoneToConversations = {};
   ConversationSummary activeConversation;
 
-  num _totalConversations = 0;
-  void set totalConversations(num v) {
+  int _totalConversations = 0;
+  void set totalConversations(int v) {
     _totalConversations = v;
     _conversationPanelTitle.text = _conversationPanelTitleText;
   }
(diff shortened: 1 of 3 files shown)


lukechurch (approved):
Nice, thanks @marianamarasoiu

webapp/lib/view.dart
```diff
@@ -815,6 +815,13 @@ class ConversationListPanelView {
   Map<String, ConversationSummary> _phoneToConversations = {};
   ConversationSummary activeConversation;
 
+  num _totalConversations = 0;
+  void set totalConversations(num v) {
```
danrubel, line 819: Should `num` be changed to `int`? Can or should this ever be a non integer?

marianamarasoiu: good point - yes, it should be an int. Fixed now, thanks!

marianamarasoiu:
thanks @danrubel and @lukechurch !

PR #429, opened 2020-08-30
Update Pipfile

marianamarasoiu:
* fix dependency vulnerability (dependabot: upgrade httplib2 to version 0.18.0 or later)
* remove un-needed libraries
* use the latest katikati-pylib

Note: accidentally developed on top of #428, this PR will need to be merged after that one, and the target branch changed to master.


lukechurch (approved):
thanks!

PR #430, opened 2020-09-03
Fix scrolling when adding new tags to the last message in the thread

### LINKED ISSUE #407: Adding new tags to the messages doesn't cause scrolling
This can make it hard to see when adding a new tag that it's happened as it often appears below the bottom of the screen

Before adding 'Does not need reply'
Screenshot 2020-07-29 at 09 08 13

After adding:
Screenshot 2020-07-29 at 09 08 17

ktkization:
Fixes #407

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -1297,12 +1297,10 @@ void setMessageTag(model.Tag tag, model.Message message, model.Conversation conv
   if (!message.tagIds.contains(tag.tagId)) {
     platform.addMessageTag(activeConversation, message, tag.tagId).then(
       (_) {
+        bool isLastMessage = conversation.messages.indexOf(message) == conversation.messages.length - 1;
         view.conversationPanelView
           .messageViewAtIndex(conversation.messages.indexOf(message))
-          .addTag(new view.MessageTagView(tag.text, tag.tagId, tagTypeToStyle(tag.type)));
-        if (conversation.messages.indexOf(message) == conversation.messages.length - 1) {
-          view.conversationPanelView.scrollToLatest();
-        }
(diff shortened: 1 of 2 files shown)


webapp/lib/controller.dart
```diff
@@ -1300,6 +1300,9 @@ void setMessageTag(model.Tag tag, model.Message message, model.Conversation conv
         view.conversationPanelView
           .messageViewAtIndex(conversation.messages.indexOf(message))
           .addTag(new view.MessageTagView(tag.text, tag.tagId, tagTypeToStyle(tag.type)));
+        if (conversation.messages.indexOf(message) == conversation.messages.length - 1) {
+          view.conversationPanelView.scrollToLatest();
```
marianamarasoiu, line 1304: can we do the action of scrolling into view directly in the `addTag` method?

webapp/lib/view.dart
```diff
@@ -364,6 +364,10 @@ class ConversationPanelView {
     _conversationWarning.title = '';
     _conversationWarning.classes.add('hidden');
   }
+
+  void scrollToLatest() {
+    _messages.scrollTop = _messages.scrollHeight;
```
marianamarasoiu, line 369: if I remember correctly, there is a `scrollIntoView` method (or something like that) on HTML elements. Could we potentially use that directly on the tag element, if it behave as expected?

webapp/lib/controller.dart
ktkization: Changed this.

webapp/lib/view.dart
ktkization: I've tied using ScrollIntoView but it doesn't seem to be working as expected.

ktkization:
Fixed by #433

PR #431, opened 2020-09-06
Unify filtering for tags with the same name

marianamarasoiu:


webapp/lib/controller.dart
```diff
@@ -1384,28 +1391,52 @@ void setMessageTag(model.Tag tag, model.Message message, model.Conversation conv
...
   // remove tags that are no longer missing their info
   var tagIdsToRemove = tagIdsWithMissingInfo.difference(newTagIdsWithMissingInfo);
   if (tagIdsToRemove.isNotEmpty) {
-    var tagsToRemove = tagIdsToTags(tagIdsToRemove, conversationTags);
-    conversationTags.removeWhere((tag) => tagsToRemove.contains(tag));
-    _removeTagsFromFilterMenu({tagsToRemove.first.group: tagsToRemove}, TagFilterType.include);
-    _removeTagsFromFilterMenu({tagsToRemove.first.group: tagsToRemove}, TagFilterType.exclude);
+    var tagsToRemove = tagIdsToTags(tagIdsToRemove, Map.fromEntries(tags.map((t) => MapEntry(t.tagId, t))));
+    tags.removeWhere((tag) => tagsToRemove.contains(tag));
+    var groupsToUpdate = Set();
```
danrubel, line 1406: You could use `= <String>{}` to provide better type validation.

danrubel (approved):
LGTM

webapp/lib/controller.dart
marianamarasoiu: good idea, thanks @danrubel !

PR #432, opened 2020-09-07
No dataset access UI

### LINKED ISSUE #397: UI for 'doesn't have access'
When a user doesn't have access to a dataset, there's currently just an endless scrolling wheel. This is causing ops problems between knowing if there's an ACL issue or a network performance issue.

Should say 'don't have access, please contact your project administrator' or the like.

ktkization:
Fixes #397

### DIFF
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -474,7 +474,10 @@ void initUI() {
         ..addAll(added)
         ..addAll(modified);
       view.conversationListSelectView.updateConversationLists(shards);
-    }, showAndLogError);
+    }, (error, stacktrace) {
+      view.conversationListPanelView.hideLoadSpinner();
+      showAndLogError(error, stacktrace);
+    });
 
   platform.listenForSystemMessages(
     (added, modified, removed) {
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -3,7 +3,7 @@ library controller;
 import 'dart:async';
 import 'dart:collection';
 
-import 'package:firebase/firebase.dart';
+import 'package:firebase/firebase.dart' show FirebaseError;
 
 import 'logger.dart';
 import 'model.dart' as model;
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -1445,7 +1445,7 @@ void showAndLogError(error, trace) {
     errMsg = "A network problem occurred: ${error.message}";
   } else if (error is Exception) {
     errMsg = "An internal error occurred: ${error.runtimeType}";
-  } else if (error is FirebaseError){
+  } else if (error is FirebaseError) {
     errMsg = "An firestore error occured: ${error.code} [${error.message}]";
     view.bannerView.showBanner("You don't have access to this dataset. Please contact your project administrator");
   } else {
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -1446,12 +1446,12 @@ void showAndLogError(error, trace) {
   String errMsg;
   if (error is PubSubException) {
     errMsg = "A network problem occurred: ${error.message}";
-  } else if (error is Exception) {
-    errMsg = "An internal error occurred: ${error.runtimeType}";
   } else if (error is FirebaseError) {
     errMsg = "An firestore error occured: ${error.code} [${error.message}]";
     view.bannerView.showBanner("You don't have access to this dataset. Please contact your project administrator");
-  } else {
+  } else if (error is Exception) {
+    errMsg = "An internal error occurred: ${error.runtimeType}";
+  }  else {
     errMsg = "$error";
   }
   command(UIAction.showSnackbar, new SnackbarData(errMsg, SnackbarNotificationType.error));
--- a/webapp/lib/controller.dart
+++ b/webapp/lib/controller.dart
@@ -378,12 +378,7 @@ void initUI() {
         view.tagPanelView.selectedGroup = selectedConversationTagsGroup;
         _populateTagPanelView(conversationTagsByGroup[selectedConversationTagsGroup], TagReceiver.Conversation);
       }
-    }, (Object error) {
-      if (error is FirebaseError) {
-        showAndLogError(error, null);
-      }
-    }
-  );
+    }, showAndLogError);
   _addDateTagToFilterMenu(TagFilterType.include);
   _addDateTagToFilterMenu(TagFilterType.exclude);
 
@@ -427,12 +422,7 @@ void initUI() {
         view.tagPanelView.selectedGroup = selectedMessageTagsGroup;
         _populateTagPanelView(messageTagsByGroup[selectedMessageTagsGroup], TagReceiver.Message);
       }
-    }, (Object error) {
-      if (error is FirebaseError) {
-        showAndLogError(error, null);
-      }
-    }
-  );
+    }, showAndLogError);
 
   if (view.urlView.shouldDisableReplies) {
     view.replyPanelView.disableReplies();
@@ -474,12 +464,7 @@ void initUI() {
         // Select the selected category in the UI and add the suggested replies for it
         view.replyPanelView.selectedCategory = selectedSuggestedRepliesCategory;
         _populateReplyPanelView(suggestedRepliesByCategory[selectedSuggestedRepliesCategory]);
-      }, (Object error) {
-        if (error is FirebaseError) {
-          showAndLogError(error, null);
-        }
-      }
-    );
+      }, showAndLogError);
   }
 
   platform.listenForConversationListShards(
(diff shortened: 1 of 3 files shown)


webapp/lib/model.g.dart
```diff
@@ -870,6 +875,7 @@ StreamSubscription<List<DocSnapshot>> listenForUpdates<T>(
     void Function(List<T> added, List<T> modified, List<T> removed) listener,
     String collectionRoot,
     T Function(DocSnapshot doc) createModel,
+    [OnErrorListener onErrorListener]
```
marianamarasoiu, line 878: since we're passing the `onErrorListener` in all calls to `listenForUpdates`, then let's not make it optional.
```suggestion
    OnErrorListener onErrorListener,
```

webapp/lib/model.g.dart
```diff
@@ -893,7 +899,9 @@ StreamSubscription<List<DocSnapshot>> listenForUpdates<T>(
       }
     }
     listener(added, modified, removed);
-  });
+  }, onError: (Object error) {
+    onErrorListener(error);
```
marianamarasoiu, line 903: if the `onErrorListener` is `null` (as it can be since it's optional in the `listenFor*` methods, then we should have a null check here:
```suggestion
    if (onErrorListener != null) onErrorListener(error);
```
(plus needs adding the stacktrace as per my other comment)

webapp/lib/controller.dart
```diff
@@ -509,6 +531,10 @@ void initUI() {
       currentUserConfig = userConfig ?? currentUserConfig;
       var newConfig = currentUserConfig.applyDefaults(defaultUserConfig);
       applyConfiguration(newConfig);
+    }, (Object error) {
+      if (error is FirebaseError) {
```
marianamarasoiu, line 535: I don't think you need to check the type of error here and only catch `FirebaseError`s - unless I'm missing something?

Also, after we change the `onErrorListener` to take the stacktrace, I think you can just pass `showAndLogError` directly, without an anonymous callback

webapp/lib/controller.dart
```diff
@@ -1448,6 +1479,9 @@ void showAndLogError(error, trace) {
     errMsg = "A network problem occurred: ${error.message}";
   } else if (error is Exception) {
     errMsg = "An internal error occurred: ${error.runtimeType}";
+  } else if (error is FirebaseError){
+    errMsg = "An firestore error occured: ${error.code} [${error.message}]";
+    view.bannerView.showBanner("You don't have access to any dataset. Please contact your project administrator");
```
marianamarasoiu, line 1484: Phrasing nit:
```suggestion
    view.bannerView.showBanner("You don't have access to this dataset. Please contact your project administrator");
```
`FirebaseError` seems like quite a generic error type - is it used only for missing permissions errors? If not, is there another way we can isolate and show this message only for missing permission errors?

webapp/lib/controller.dart
```diff
@@ -1448,6 +1479,9 @@ void showAndLogError(error, trace) {
     errMsg = "A network problem occurred: ${error.message}";
   } else if (error is Exception) {
     errMsg = "An internal error occurred: ${error.runtimeType}";
+  } else if (error is FirebaseError){
```
marianamarasoiu, line 1482: nit: 
```suggestion
  } else if (error is FirebaseError) {
```

webapp/lib/controller.dart
```diff
@@ -3,6 +3,8 @@ library controller;
 import 'dart:async';
 import 'dart:collection';
 
+import 'package:firebase/firebase.dart';
```
marianamarasoiu, line 6: Nit: 
```suggestion
import 'package:firebase/firebase.dart' show FirebaseError;
```
(let's narrow down the import, since we shouldn't really use the firebase package in this file, but we now need to import it for the firebase error - this is similar to what we do for the `PubSubException`)

webapp/lib/model.g.dart
marianamarasoiu: Actually can you check if we can just pass the `onErrorListener` directly rather than wrapping it in the anonymous callback? something like `onError: onErrorListener`. And also what happens if it's `null` - I would hope that it just throws the error in the console as if there was no listener?

marianamarasoiu (commented):
thanks @ktkization - just a few nits and a couple of questions

webapp/lib/controller.dart
ktkization: It seems like it's a generic error and not specific to permission error as seen here [doc1](https://pub.dev/documentation/firebase/latest/firebase/FirebaseError-class.html), [doc2](https://firebase.google.com/docs/reference/js/firebase.FirebaseError). We'd probably have to come up with a better way.

webapp/lib/controller.dart
ktkization: Nice suggestion. TIL

webapp/lib/controller.dart
ktkization: Right. I've passed `showAndLogError` instead directly.

webapp/lib/model.g.dart
ktkization: Yes, we can be able to pass the `onErrorListener` directly. Nothing happens if it's `null`. I'm guessing it's because it only gets executed when there is an actual error.

webapp/lib/model.g.dart
```diff
@@ -653,6 +654,10 @@ typedef UserConfigurationCollectionListener = void Function(
   List<UserConfiguration> removed,
 );
 
+typedef OnErrorListener = void Function(
+  Object error,
+);
```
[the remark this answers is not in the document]
ktkization: Yes. I have done that.

webapp/lib/controller.dart
ktkization: What do you suggest?

webapp/lib/controller.dart
ktkization: Fixed

webapp/lib/controller.dart
```diff
@@ -475,8 +474,7 @@ void initUI() {
         ..addAll(added)
         ..addAll(modified);
       view.conversationListSelectView.updateConversationLists(shards);
-    }
-  );
+    }, showAndLogError);
```
marianamarasoiu, line 477: in testing I saw that the load spinner still appears, which might give the impression to the user that something is still loading. I think something like this will do the trick to hide it:

```suggestion
    }, (error, stacktrace) {
      view.conversationListPanelView.hideLoadSpinner();
      showAndLogError(error, stacktrace);
    });
```

marianamarasoiu (approved):
LGTM thanks @ktkization !

@danrubel can you also please have a look at this PR, in particular at the changes to model.g.dart which I think we'll need to transfer into the code generator? I'm happy to do that if these changes look ok to you. thanks!

webapp/lib/controller.dart
```diff
@@ -1448,6 +1445,9 @@ void showAndLogError(error, trace) {
     errMsg = "A network problem occurred: ${error.message}";
   } else if (error is Exception) {
     errMsg = "An internal error occurred: ${error.runtimeType}";
+  } else if (error is FirebaseError) {
```
danrubel, line 1448: Since `FirebaseError` is an abstract class, then it may now or in the future be a subclass of `Exception`. If that is or ever becomes the case, then this `else if` clause will never be triggered because the `else if` clause above it will be true. To address this, I recommend putting the `else if (error is FirebaseError) {` clause before the `else if (error is Exception) {` clause

webapp/lib/model.g.dart
```diff
@@ -3,6 +3,7 @@
 
 import 'dart:async';
 
+
```
danrubel, line 6: nit: Unnecessary blank line ?

webapp/lib/model.g.dart
```diff
@@ -653,6 +654,11 @@ typedef UserConfigurationCollectionListener = void Function(
   List<UserConfiguration> removed,
 );
 
+typedef OnErrorListener = void Function(
+  Object error,
+  StackTrace stackTrace
+);
```
danrubel, line 660: I like these model changes! Nice work!
Once this lands, I can update model codegen to include these changes.

danrubel (approved):
LGTM with the following changes...

webapp/lib/controller.dart
ktkization: Fixed

webapp/lib/controller.dart
ktkization: Good catch. Refactored it.

webapp/lib/model.g.dart
ktkization: Removed

webapp/lib/model.g.dart
ktkization: cool

ktkization:
@marianamarasoiu Can you please help sort out the merge conflict when you're merging. I don't want to mistakenly leave out something.

marianamarasoiu (approved):
great, thanks @ktkization for the changes and @danrubel for the review!

@ktkization - yup, I'll take over merging and handling the conflicts

marianamarasoiu:
For some reason the build is failing for this PR, but it doesn't look like it's caused by the changes to this code. It seems instead that the cause is that our pubspec is now behind the `google/dart:latest` container we use for building nook on GitHub. I'll merge this code since it works fine locally, and we'll update the pubspec separately.

PR #434, opened 2020-09-09
Use specific 2.5.9 version of webdev

marianamarasoiu:
`webdev 2.6` is the latest version of `webdev`, but it requires at least `build_web_compilers` version `2.12.0-dev.1`, which isn't production-ready yet, so `pub get` doesn't download it.

When `build_web_compilers 2.12.0` gets released, we can roll back this change.


danrubel (approved):
LGTM!

PR #435, opened 2020-09-13
Add multiple reply message packs

marianamarasoiu:
Deployed to [nook-dev](https://nook-development.firebaseapp.com/?) for testing.

Next step is to add `send_messages_to_ids` action to the pubsub handler, and then transition Nook to only using that one instead of `send_to_multi_ids`.


danrubel (approved):
LGTM!

PR #436, opened 2020-09-15
Use single reply API for reply groups

marianamarasoiu:


marianamarasoiu:
This was a temporary fix until we landed multi replies into the infrastructure, we can now close this without merging.

PR #439, opened 2020-09-27
Fix last inbound turn tag filter not loading correctly

marianamarasoiu:
The last-inbound-turn tags are applied to the message not conversation - this PR loads them correctly from the message tags collection and refreshes the UI when the message tags have been downloaded from firebase


danrubel (approved):
LGTM

PR #440, opened 2020-09-27
Fix single shard projects not loading any data

marianamarasoiu:
This was caused by the previous PR #417 that removed the shard-selection logic from view (which treated single-shard selectors differently to multi-shards), but then I didn't introduce it back in the controller.


danrubel (approved):
Good fix. Thank you!

PR #457, opened 2021-02-03
Integrate auth and snackbar to nook

elayabharath:


webapp/lib/platform.dart
```diff
@@ -6,7 +6,8 @@ import 'controller.dart' as controller;
 import 'logger.dart';
 import 'model.dart';
 import 'model_firebase.dart';
-import 'platform_constants.dart' as platform_constants;
+import 'package:katikati_ui_lib/components/platform/platform_constants.dart' as platform_constants;
+import 'package:katikati_ui_lib/components/platform/platform.dart' as platform;
```
marianamarasoiu, line 10: I think we can also remove a few more methods from lines 163 to 180 since you've also moved them to this file. And then here you can also export the library so the other files in the project who need to call those functions, don't also need to import the kk_ui_lib package.
```suggestion
import 'package:katikati_ui_lib/components/platform/platform.dart' as platform;
export 'package:katikati_ui_lib/components/platform/platform.dart';
```

marianamarasoiu (approved):
thanks @elayabharath - LGTM with just a suggestion to also use the sign in/out methods from the library, and how to export them for the rest of the app to use as needed

webapp/lib/platform.dart
elayabharath: Ah yes, sorry my bad in overlooking this. Fixed along with the signin method

PR #459, opened 2021-02-04
Prepare Nook UI for demo

marianamarasoiu:
* add Katikati and IFRC logos
* remove cohort messages filtering
* replace suggested responses with "standard messages" 
* removed multiple options for sign in and left only lark.systems

Note: this is merging into a `main-demo` branch


marianamarasoiu:
TBR

PR #461, opened 2021-02-07
Report user presence

lukechurch:
@marianamarasoiu - this adds a simple transitory user presence system for reporting which conversation a user had opened.


marianamarasoiu (approved):
LGTM, thanks!

PR #462, opened 2021-02-07
Integrated auth login page from katikati_ui_lib

elayabharath:
- Login page from katikati_ui_lib
- Added common styles via global.css and removed the rest of the css
- Removed unused snackbar.css
- `hidden` class has been moved to utils


webapp/web/styles.css
```diff
@@ -433,31 +433,7 @@ main {
...
-
-img.partner-logo {
-    height: 60px;
-    margin: 16px;
-}
-
 img.partner-logo--unicef {
     height: 40px;
     padding-bottom: 3px;
 }
```
marianamarasoiu, line 439: It would be good to also remove this, it's now obsolete code
```suggestion
```

marianamarasoiu (approved):
this is great, thanks @elayabharath ! just another suggestion for tidying up another piece of code that we no longer need.

webapp/lib/view.dart
```diff
@@ -36,7 +37,13 @@ void init() {
   replyPanelView = new ReplyPanelView();
   tagPanelView = new TagPanelView();
   authHeaderView = new AuthHeaderView();
-  authMainView = new AuthMainView();
+  authMainView = new AuthMainView(
+    "assets/africas-voices-logo.svg",
+    "Welcome to Nook.", 
+    "Sign in to Nook where you can manage SMS conversations.",
+    [SignInDomain.avf, SignInDomain.lark, SignInDomain.ucam, SignInDomain.gmail],
```
marianamarasoiu, line 44: I think the others snuck in from one of the Cambridge collaboration projects - as the default we should only have AVF and Lark to match the firestore rules.
```suggestion
    [SignInDomain.avf, SignInDomain.lark],
```

elayabharath:
Thanks @marianamarasoiu Didn't realise they weren't actually used, just tested by opening the google's & cambridge's auth popups.

PR #464, opened 2021-02-09
Show other users' activity

marianamarasoiu:
* show small coloured dots next to the conversation ID in the list for each logged in user
* show small coloured dots in the header next to the logged in user info. Clicking one of those dots will jump to the conversation that that user is looking at.

Towards https://github.com/larksystems/Katikati-Core/issues/417


danrubel (approved):
LGTM!

PR #465, opened 2021-02-10
Add validation for the cloud function URLs

marianamarasoiu:
Fixes https://github.com/larksystems/Katikati-Core/issues/403


danrubel (approved):
Good idea. LGTM!

PR #472, opened 2021-02-27
Remove disable replies feature from the URL parameter

elayabharath:
+ Removed the url parameter to disable replies feature
+ Set the config `replies_panel_visibility` to false for a specific account in firebase to see if the suggested panels is hidden


marianamarasoiu (approved):
thanks @elayabharath !

PR #473, opened 2021-02-28
Integrate nook configurator and nook

marianamarasoiu:
(Apologies in advance for the huge diff. The main functionality has already been reviewed both in nook and the configurator, so if possible, you should aim to focus on the controller/view hierarchy and file layout)

The main architectural change is to turn the controller, the platform and the view into objects that get instantiated on page load (first the controller, which then instantiates the platform and view).

Each "page" has it's own `.html` file and it's own `main` function, which calls the right controller. Switching between different "pages" (e.g. from nook to the main page) loads a new html page, rather than loading it all in the same app. This makes the logic of swapping views around much simpler (we don't, we just get a new page), but depending on where the UI is heading, it may not be the right choice. However, I think switching to the other architecture if needs be, will be easier from the current setup.

The key files are as follows:
- controller.dart (the main controller interface and base functionality)
- view.dart (the main view interface and base functionality)

Each of the implemented pages has their own subclassed controller and view, and map to the HTML as follows:

- homepage/ --> web/index.html
- nook/ --> web/converse/index.html
- configurator/tags/ --> web/configure/tags.html
- configurator/standard_messages/ --> web/configure/messages.html

There's an issue at the moment with `webdev serve` as it doesn't correctly serve `converse/index.html` (the expected behaviour would be to serve it as `configure/`, but only the full path works). I am investigating potential fixes and will update on what I find, otherwise the fix would be to have nook show under the full `converse/index.html` path


elayabharath (approved):
Mostly skimmed through the files: I really like the split of the controller & view based on the pages, and further splitting the controller to utils / view helpers!

marianamarasoiu:
Thanks both - and sorry again for the un-reviewable diff! Merging now and will do more testing and bug fixing on top of master.

PR #474, opened 2021-03-02
Basic test scenarios

elayabharath:


webapp/documentation/test_scenarios.md
```diff
@@ -0,0 +1,24 @@
+# Navbar
+  - Signed in state
+    - [ ] Auth: Nav bar should contain user auth (profile photo, name, and sign out). Sign out should kick the user to the authentication page.
+    - [ ] User presence: when another user concurrently uses nook, the person indicator (colored dots) appears. Clicking on them should take to the respective conversation.
+        - Test by adding a conversation id and the time stamp under `user_presence > user email`
+    - [ ] Conversation list panel: A dropdown appears when there are more than 1 conversation shards present.
+	- Test by adding a conversation name in firebase under `nook_conversation_shards > shard-n > name (field), conversations (collection)`
```
marianamarasoiu, line 7: There's a few more steps to do here, as the number of shards is given by a field the first:

```suggestion
	- If the test project doesn't already have 2 shards, test by setting the field `nook_conversation_shards > shard-0 > num_shards: 2` and adding a second conversation shard with a name `nook_conversation_shards > shard-1 > name: "Conversation list #2"`
```

webapp/documentation/test_scenarios.md
```diff
@@ -0,0 +1,24 @@
...
+    - [ ] Style: Nav bar shouldn't hide any of the content below
+
+  - Signed out state
+    - [ ] Navbar should contain only sign-in button, should log in via the first domain name
```
marianamarasoiu, line 11: I think it might look better not to be a navbar at all in sign-out state, just the login page - what do you think?

webapp/documentation/test_scenarios.md
```diff
@@ -0,0 +1,24 @@
...
+
+# Banner
+ - [ ] If the user doesn't have access to the data set, a banner should appear at the top
```
marianamarasoiu, line 14:
```suggestion
 - [ ] If the user doesn't have access to the data set, a banner should appear at the top
   - [ ] Test by removing your email address from the `users` collection in firebase
```

webapp/documentation/test_scenarios.md
```diff
@@ -0,0 +1,24 @@
...
+
+# User permissions
+ - Toggle the following flag to see the nook UI reacting to it
+   - [ ] `replies_panel_visibility` should toggle suggested replies section on the right. If the `edit_notes_enabled` is present, the notes text area should take up the whole height of the sidebar.
```
marianamarasoiu, line 18:
```suggestion
   - [ ] `replies_panel_visibility` should toggle suggested replies section on the right. If the `edit_notes_enabled` is `true`, the notes text area should take up the whole height of the sidebar.
```

webapp/documentation/test_scenarios.md
```diff
@@ -0,0 +1,24 @@
...
+   - [ ] `edit_notes_enabled` should toggle notes text area.
```
marianamarasoiu, line 19:
```suggestion
   - [ ] `edit_notes_enabled` should toggle the notes text area.
```

webapp/documentation/test_scenarios.md
```diff
@@ -0,0 +1,24 @@
...
+   - [ ] `replies_panel_visibility && edit_notes_enabled` are disabled, the whole sidebar panel should not be visible.
```
marianamarasoiu, line 20: nit: just a bit more explicit as to what disabled means
```suggestion
   - [ ] if both `replies_panel_visibility` and `edit_notes_enabled` are `false`, the whole sidebar panel should not be visible.
```

marianamarasoiu (approved):
this is excellent, thanks @elayabharath - just some nits to have a bit more detailed instructions, but LGTM otherwise!

webapp/documentation/test_scenarios.md
marianamarasoiu: (this is what I ended up doing in the merge, but would appreciate your thoughts on it, am happy to revert the change!)

elayabharath: Yep that makes sense! I just noticed it along with the comment

elayabharath:
Moved this to main since the history had changed

PR #476, opened 2021-03-06
Selective toggle for replies / notes panel

elayabharath:
+ Hides / shows entire panel only when either notes / replies panel is enabled
+ Clears the panel title before appending


marianamarasoiu (approved):
LGTM!

PR #478, opened 2021-03-07
Move User presence to be generated by modelgen

elayabharath:
+ Update model path
+ Made sure there is no diff in model.g.dart (except the commented out log at L315) against the script at https://github.com/larksystems/Infrastructure/pull/43


marianamarasoiu (approved):
very nice, thanks @elayabharath !

PR #479, opened 2021-03-10
Fix local variable declaration shadowing class attribute

marianamarasoiu:
I think this fix slipped by in the repo restructuring from a few weeks ago, so adding it now


elayabharath (approved):

elayabharath:
For some reason I thought I had fixed this, my bad i missed out!

marianamarasoiu:
I thought so too, I remember discussing it with you and then reviewing, so no idea why it wasn't in, hence I thought I might have missed it when I brought in the configurator if I accidentally didn't work from head.. No worries either way - we know now what the fix needed to be. And thanks for the quick review!

PR #480, opened 2021-03-11
Integrate model from katikati ui lib

elayabharath:
+ Tested with editing the tags, messages, selecting the conversations from the ui, applying tags. Haven't seen any issues so far!


marianamarasoiu (approved):
LGTM, thanks EB! 🚀

PR #481, opened 2021-03-11
Back link to the main config page from converse UI

elayabharath:
Ref: https://github.com/larksystems/Katikati-Core/issues/412

+ Moved the navbar to use grid so that the conversation selector, back to main config page, user presence can all co-exist
  - We may want to move the conversation selector to the top of conversations panel from the navbar

+ Minor clean up of spacing in the home page
  - The items under the title are closer to the its title than the neighbouring title
  - Buttons are of same width

### DIFF
--- a/webapp/lib/app/nook/view.dart
+++ b/webapp/lib/app/nook/view.dart
@@ -69,9 +69,15 @@ class NookPageView extends PageView {
       showNormalStatus('signed in: ${latestCommitHash.substring(0, 8)}...');
     }, onError: (_) { /* Do nothing */ });
 
+    var backPageLink = new Element.a()
+      ..classes.add('configuration-view__back-link')
+      ..text = 'Back to the main configuration page'
+      ..onClick.listen((event) => controller.routeToPage(Page.homepage));
+
     navHeaderView.navContent = new DivElement()
+      ..classes.add('nav-content__grid-wrapper')
       ..append(conversationListSelectView.panel)
-      ..append(new DivElement()..classes.add('flex-fill-gap'))
+      ..append(new DivElement()..append(backPageLink))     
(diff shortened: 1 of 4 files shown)


webapp/lib/app/nook/view.dart
```diff
@@ -69,9 +69,15 @@ class NookPageView extends PageView {
       showNormalStatus('signed in: ${latestCommitHash.substring(0, 8)}...');
     }, onError: (_) { /* Do nothing */ });
 
+    var backPageLink = new Element.a()
+      ..classes.add('configuration-view__back-link')
```
marianamarasoiu, line 73: the css for this class is defined only for the `configure` pages, not for converse, so the css for it won't actually apply in Nook.. I think this is a good candidate to generalise and move it into a top-level css file?

marianamarasoiu (commented):
I like the direction where this is heading, and thanks for making the buttons on the first page the same size! 

I agree that it might be a good idea to have the conversation list selector in the conversation panel rather than in the navigation bar, but having the text to go back to the main page in the middle feels a bit odd at the moment. Multi-shard projects are pretty rare, so let's:
* have by default the text for "Back to..." be aligned left. So when there's only one conversation list and the selector isn't visible, it appears fully to the left of the page.
* when the selector is visible it should appear to the right of it (with a bit of distance), or in the middle, whichever is easiest, as this stage will be temporary

Could you also make sure that the other user presence indicators always appear aligned to the right, next to the logged in user image? thanks!

marianamarasoiu:
@elayabharath can we close this PR as the back link was added in #495 instead?

elayabharath:
Ah yes this can be closed, will do!

PR #482, opened 2021-03-11
Replace × with "Close" for the sample messages tooltip

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/449


marianamarasoiu (approved):
nice, thanks EB!

PR #484, opened 2021-03-13
Configure messages page: Moved delete button from the left to the right side

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/450


marianamarasoiu (approved):
LGTM!

PR #487, opened 2021-03-13
Handle tag modification as removal and addition

marianamarasoiu:
This allows better handling when the group of the tag changes, which currently fails with an "item not in list" error in view.dart on line 1150 as the previous group is lost, and the new group doesn't have that element.


danrubel (approved):
LGTM!

elayabharath (approved):

PR #488, opened 2021-03-19
Add ability to delete suggested replies

marianamarasoiu:


webapp/lib/app/configurator/standard_messages/controller.dart
```diff
@@ -111,22 +110,21 @@ class MessagesConfiguratorController extends ConfiguratorController {
...
+        editedStandardMessages[standardMessage.docId] = standardMessage;
         break;
 
       case MessagesConfigAction.removeStandardMessage:
         StandardMessageData messageData = data;
         var standardMessage = standardMessagesManager.getStandardMessageById(messageData.id);
         standardMessagesManager.removeStandardMessage(standardMessage);
         _view.groups[standardMessage.groupId].removeMessage(standardMessage.suggestedReplyId);
-        // TODO: queue suggested messages for removal once the backend infrastructure can handle removing them
+        removedStandardMessages[standardMessage.suggestedReplyId] = standardMessage;
```
danrubel, line 121: similar to line 143 below, should we add the following line above?
`editedStandardMessages.remove(standardMessage.suggestedReplyId);`

danrubel (approved):
LGTM!

webapp/lib/app/configurator/standard_messages/controller.dart
marianamarasoiu: good catch, thanks!

PR #490, opened 2021-03-30
Remove mark unread button

marianamarasoiu:
The read/unread functionality isn't used at the moment since it doesn't work well when multiple people are using Nook and so it's currently disabled [from the engine](https://github.com/larksystems/katikati_lib.dart/blob/0890213290a35cdfe3e281917fed167b291ae9b7/lib/perspectives/export_to_firebase.dart#L146). This PR just clears up the UI for now, until we rethink how/if we should better design this and show it to users.

cc @elayabharath @lukechurch


danrubel (approved):
LGTM

PR #491, opened 2021-03-30
Remove button to hide demog tags

marianamarasoiu:
The button is no longer needed since tags can now be grouped

cc @elayabharath @lukechurch


danrubel (approved):
LGTM

PR #495, opened 2021-04-03
Integrate back page style from katikati_ui_lib

elayabharath:
+ Verified that the back page link is before the conversation selector, and the user presence is aligned to the right


marianamarasoiu (approved):
LGTM!

PR #496, opened 2021-04-03
Messages refactor

elayabharath:
+ Made the design similar to the tags page
+ The collapse and expand works on the group level
+ Cleaned up layout to have standard padding, sizes
+ Save config floats over the messages when the form is dirty
+ When removing a message group, it expands to show all the messages that would be affected


marianamarasoiu (approved):
LGTM, thanks @elayabharath !

PR #499, opened 2021-04-11
Layout for nook

elayabharath:
+ Independently scrollable panels, with the existing functionality
+ Selected messages styling
+ Display none for hiding the after date filter since it was adding a few pixels at the bottom
+ Updated a inline divs to spans

+ Pending on the selective toggle for the tags panel - this seems like a lot to take on in this PR, hence will push as separate PR


marianamarasoiu (approved):
LGTM! thanks @elayabharath

PR #500, opened 2021-04-14
Enable save tag config button on changes

marianamarasoiu:
The view starts with a disabled save config button, and with the tags and messages sharing that code, but only being supported by the messages page, the tags config page was never enabling that button, and so you couldn't save.

Also:
* resets the bool keeping track of whether the form is dirty or not after a successful save
* marks form dirty when a message group is removed


lukechurch (approved):
Thanks @marianamarasoiu!

PR #501, opened 2021-04-19
Remove handling of separate message tags in Nook

marianamarasoiu:
First step towards https://github.com/larksystems/Katikati-Core/issues/496

Next step will be to rename `[...]conversationTags` as just `[...]tags`


danrubel (approved):
LGTM

PR #506, opened 2021-05-04
Move edited/deleted tag tracking in the tag manager

marianamarasoiu:
Also includes a bug fix for the case where you would add a new group, rename it, then add a new tag - which would wrongly use the previous group name (the change for this is in `view.dart` only)

cc @lukechurch


danrubel (approved):
LGTM

PR #508, opened 2021-05-05
Add suggested messages in a conversation

marianamarasoiu:
This is still a prototype and in need of a bit more testing. Long term, we should be using the components that are being developed in kk_ui_lib, but for now, this just adjusts the existing UI elements slightly to accommodate a suggested messages panel below the message text box.

Relies on https://github.com/larksystems/katikati_ui_lib.dart/pull/34, which will need to be merged first, and the pubspec ref updated here.

cc @lukechurch

### DIFF
--- a/webapp/lib/app/nook/controller.dart
+++ b/webapp/lib/app/nook/controller.dart
@@ -1478,7 +1478,7 @@ class NookController extends Controller {
     }
 
     log.verbose('Sending ${textReplies.length} replies "${repliesStr}" to conversation ${conversation.docId}');
-    platform.sendMessages(conversation.docId, textReplies, wasSuggested: true, onError: (error) {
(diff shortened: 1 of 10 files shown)


webapp/lib/app/nook/controller.dart
```diff
@@ -1459,7 +1478,7 @@ class NookController extends Controller {
     }
 
     log.verbose('Sending ${textReplies.length} replies "${repliesStr}" to conversation ${conversation.docId}');
-    platform.sendMessages(conversation.docId, textReplies, onError: (error) {
+    platform.sendMessages(conversation.docId, textReplies, wasSuggested: true, onError: (error) {
```
marianamarasoiu, line 1481:
```suggestion
    platform.sendMessages(conversation.docId, textReplies, wasSuggested: wasSuggested, onError: (error) {
```

elayabharath (approved):

PR #509, opened 2021-05-09
Integrate send message along with the 160 character limit

elayabharath:


marianamarasoiu (approved):
LGTM thanks @elayabharath ! just a couple of nits below

PR #510, opened 2021-05-09
Integrate button links in the navbar instead of having back button to configure page

elayabharath:


webapp/lib/app/utils.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:katikati_ui_lib/components/nav/button_links.dart';
+
+var navLinks = [
+  Link("Dashboard", "/"),
+  Link("Conversations", "/converse/index.html"),
+  Link("Messages", "/configure/messages.html"),
+  Link("Tags", "/configure/tags.html")
```
marianamarasoiu, line 7: can we add a link to the explorer here as well?
```suggestion
  Link("Tags", "/configure/tags.html"),
  Link("Explore", /explore"),
```

marianamarasoiu (approved):
LGTM, thanks EB! could you also prepare a PR that does the same thing for the explorer UI? (thanks in advance!)

PR #511, opened 2021-05-12
Integrate conversation item

elayabharath:


webapp/lib/app/nook/view.dart
```diff
@@ -1222,56 +1225,36 @@ class ConversationSummary with LazyListViewItem, UserPresenceIndicator {
...
+      ..onUncheck.listen((_) {
+        _view.appController.command(UIAction.deselectConversation, new ConversationData(deidentifiedPhoneNumber));
+      })
+      ..onSelect.listen((_) {
+        _view.appController.command(UIAction.showConversation, new ConversationData(deidentifiedPhoneNumber));
+      });
+
+    if (_warning) {
+      _conversationItem.setWarnings(Set.from([ConversationWarning.notInFilterResults]));
     }
```
marianamarasoiu, line 1253: As per my comment above, any reason why we're removing this?

marianamarasoiu (commented):
hi EB - this broadly looks good, but just a couple of questions for you on the removal of the presence indicator

marianamarasoiu (approved):
just saw the #512 , so my questions here are no longer an issue. LGTM, thanks!

PR #515, opened 2021-05-31
add pubspec lock file

danrubel:
add pubspec.lock so everyone is using the same set of dependent packages


marianamarasoiu (approved):
LGTM, thanks!

PR #520, opened 2021-06-10
Integrate tag for conversation summary

elayabharath:


webapp/lib/app/nook/view.dart
```diff
@@ -760,34 +760,27 @@ class SuggestedMessageTagView extends TagView with AutomaticSuggestionIndicator
...
+    };
 
-    var confirmButton = new SpanElement()
-      ..classes.add('tag__confirm')
-      ..classes.add('btn')
-      ..classes.add('btn--hover-only')
-      ..onClick.listen((_) {
-        DivElement messageSummary = getAncestors(tag).firstWhere((e) => e.classes.contains('conversation-summary'));
+    onAccept = () {
+      super.markPending(true);
```
marianamarasoiu, line 780: I don't think you need `super.` here because you're not overwriting the method, it should just work without it:

```suggestion
      markPending(true);
```

webapp/lib/app/nook/view.dart
```diff
@@ -760,34 +760,27 @@ class SuggestedMessageTagView extends TagView with AutomaticSuggestionIndicator
...
+      DivElement messageSummary = getAncestors(renderElement).firstWhere((e) => e.classes.contains('conversation-summary'));
         _view.appController.command(UIAction.confirmConversationTag, new ConversationTagData(tagId, messageSummary.dataset['id']));
```
marianamarasoiu, line 782: I think the indentation for this line should be adjusted slightly to match the line above?
```suggestion
      _view.appController.command(UIAction.confirmConversationTag, new ConversationTagData(tagId, messageSummary.dataset['id']));
```

marianamarasoiu (approved):
LGTM, just a couple of nits, thanks @elayabharath !

PR #521, opened 2021-06-13
Integrate KK tag to configure tags

elayabharath:


webapp/lib/app/configurator/tags/controller_view_helper.dart
```diff
@@ -1,15 +1,24 @@
 part of controller;
 
-void _addTagsToView(Map<String, List<model.Tag>> tagsByCategory) {
+void _addTagsToView(Map<String, List<model.Tag>> tagsByCategory, {bool editable = false}) {
```
marianamarasoiu, line 3: As you mentioned earlier that `editable` isn't the right name and that the API might need changing, perhaps we can an anticipate that change in this PR by using something like `startEditing` instead?

marianamarasoiu (approved):
LGTM with a small comment, thanks @elayabharath !

PR #524, opened 2021-06-13
Bottom margins for tags collection

elayabharath:


marianamarasoiu (commented):
would it make sense to have margin on the `tag` class in the ui library css so we don't have to do it for (almost) every use case?  If we do that, which is the case that won't work?

PR #526, opened 2021-06-16
Remove unused tag, kk namespace for tag

elayabharath:


marianamarasoiu (approved):
LGTM!

Would you mind also taking a look at cleaning up the css for tags in Nook since it's now all coming from the ui library? thanks!

PR #528, opened 2021-06-19
Add button for sorting the conversation list

marianamarasoiu:
The conversation list can now be sorted:
* alphabetically by conversation ID
* by the timestamp of the last incoming message

Also makes the sorting of the conversation list live - ie. it will be updated on each new incoming message

Closes https://github.com/larksystems/Katikati-Core/issues/524

Note to self: dependent on https://github.com/larksystems/katikati_common_lib/pull/52 so that one needs to be merged first and then the pubspec files here need to be updated with the merge SHA


elayabharath (approved):
Looks good!

PR #529, opened 2021-06-24
Integrate accordion, editable text for messages config page

elayabharath:
Pending:
1. Rename of the group name not working


marianamarasoiu (approved):
LGTM, thanks @elayabharath !

PR #531, opened 2021-07-03
Integrate accordion for filter tags menu, tags config page

elayabharath:


webapp/lib/app/configurator/tags/view.dart
```diff
@@ -55,76 +56,74 @@ class TagsConfigurationPageView extends ConfigurationPageView {
...
+      renderElement.classes.toggle('dropzone--active', true);
+    });
+    tagsDropzone.onDragLeave.listen((event) {
+      renderElement.classes.toggle('dropzone--active', false);
+    });
     tagsDropzone.onDrop.listen((event) {
       var tag = event.draggableElement;
       var tagId = tag.dataset['id'];
       var groupId = tag.dataset['group-id'];
+      expand();
```
marianamarasoiu, line 103: can we change the logic for expanding the accordion to be done on drag over rather than on drop? so:
* onDragOver: if not expanded, expand and remember that it was collapsed
* onDragLeave: if it was collapsed before drag, collapse it again

webapp/lib/app/nook/view.dart
```diff
@@ -1135,52 +1135,47 @@ class ConversationFilter {
...
-
-      return newContainer;
-    }).append(tag.renderElement);
-    bool wasHidden = _tagGroupsContainers[category].classes.remove('hidden'); // briefly override any display settings to make sure we can compute getBoundingClientRect()
-    List<num> widths = _tagGroupsContainers[category].querySelectorAll('.tag__text').toList().map((e) => e.getBoundingClientRect().width).toList();
-    _tagGroupsContainers[category].classes.toggle('hidden', wasHidden); // clear inline display settings
-    num avgGridWidth = widths.fold(0, (previousValue, width) => previousValue + width);
-    avgGridWidth = avgGridWidth / widths.length;
+
+    if(_tagAccordion.queryItem(category) == null) {
```
marianamarasoiu, line 1144: nit:
```suggestion
    if (_tagAccordion.queryItem(category) == null) {
```

webapp/lib/app/nook/view.dart
```diff
@@ -1135,52 +1135,47 @@ class ConversationFilter {
...
+      _tagAccordion.appendItem(tagGroup);
+    }
+
+    var currentAccordion = _tagAccordion.queryItem(category);
+    currentAccordion.bodyElement.append(tag.renderElement);
+
+    // todo: potentially move this computation to onExpand
+    // recompute the grid
+    var wasHidden = !currentAccordion.isOpen;
+    if(wasHidden) {
```
marianamarasoiu, line 1155: nit:
```suggestion
    if (wasHidden) {
```

webapp/lib/app/nook/view.dart
```diff
@@ -1135,52 +1135,47 @@ class ConversationFilter {
...
+      currentAccordion.expand();
+    }
+    List<num> widths = currentAccordion.bodyElement.querySelectorAll('.tag__text').toList().map((e) => e.getBoundingClientRect().width).toList();
+    if(wasHidden) {
```
marianamarasoiu, line 1159: nit:
```suggestion
    if (wasHidden) {
```

webapp/lib/app/nook/view.dart
```diff
@@ -1135,52 +1135,47 @@ class ConversationFilter {
...
+      currentAccordion.collapse();
+    }
+    if(widths.isEmpty) {
```
marianamarasoiu, line 1162: nit:
```suggestion
    if (widths.isEmpty) {
```

webapp/lib/app/nook/view.dart
```diff
@@ -1135,52 +1135,47 @@ class ConversationFilter {
...
+    currentAccordion.bodyElement.style.setProperty('grid-template-columns', 'repeat(auto-fill, ${columnWidth}px)');
   }
 
   void removeMenuTag(FilterMenuTagView tag, String category) {
-    int index = _tagGroups[category].indexWhere((t) => t.renderElement.dataset["id"] == tag.renderElement.dataset["id"]);
-    _tagGroups[category][index].renderElement.remove();
-    _tagGroups[category].removeAt(index);
+    _tagAccordion.items.forEach((accordion) {
+      accordion.bodyElement.children.removeWhere((element) => element.dataset["id"] == tag.renderElement.dataset["id"]);
+    });
```
marianamarasoiu, line 1178: This doesn't do the same thing that the older code does - it should be only the tag from the accordion with the id equal to `category` that gets removed, not from all accordion items.

marianamarasoiu (approved):
LGTM with a couple of comments and a few nits, thanks @elayabharath !

webapp/lib/app/configurator/tags/view.dart
elayabharath: Oh I had originally tried this but it is quite janky
If there are say 10 groups A, B, C ... with everything collapsed at the start
Say we drag a tag from A to F

Once out of A, B opens. Once out of B, B is closed but the hover position could be such that it is over E that the group E opens. All this happens quite fast that sometimes it is confusing as to what group you are hovering over. 

Please let me know if there is a better way to do this?

marianamarasoiu: hmm, yeah, I can see how that would happen. let's keep the current behaviour as is, and look to implement the target behaviour later on, as I think it'll be quite complex (e.g. adjusting the page scroll when folding so that the page appears stable relative to the cursor position on the page). I've opened an issue for tracking this: https://github.com/larksystems/Katikati-Core/issues/548

marianamarasoiu (approved):
thanks @elayabharath - LGTM, we'll resolve the remaining comment later on as it's not a priority, and will likely take quite a bit of work

webapp/lib/app/nook/view.dart
elayabharath: Fixed this!

elayabharath:
Thanks @marianamarasoiu I have added a few fixes as well to this - looks like the model has changed, so to make it work with the latest version of common_lib had to add it in.

PR #533, opened 2021-07-12
Refix the model that has been reverted in common lib

elayabharath:
https://github.com/larksystems/katikati_common_lib/commit/7d831b8a6da0fc99311d385731ce56375c883242


marianamarasoiu:
@danrubel can I ask you to take a look at this? Is the engine exporting `TagType.normal` or `TagType.Normal`?

danrubel:
@marianamarasoiu Yes this looks correct as I have reverted my earlier work for now
         (see https://github.com/larksystems/katikati_common_lib/pull/62)

@elayabharath Sorry about forgetting to revert nook and thank you for fixing!

danrubel (approved):
LGTM!

marianamarasoiu (approved):
Ah, got it, sorry for the noise, I forgot that `TagType` is a class not an enum, and its values are explicitly matched to lowercase strings.

Thanks @danrubel  and @elayabharath - LGTM from me as well then!

PR #534, opened 2021-07-12
Remove unused css under config messages, tags page

elayabharath:


webapp/web/configure/buttons.css
```diff
@@ -1,64 +1,5 @@
 @charset "utf-8";
```
marianamarasoiu, line 1: a few of the classes in this file marked for deletion are still used, e.g. `button--confirm`, `button--contained`, `button--add` etc. I suggest we keep this file as is for now, then we transition to using the button class from the common library and delete it then.

marianamarasoiu (approved):
thanks EB, LGTM except for the `buttons.css` - I suggest we revert that back, and remove it in one go when we switch from the local implementation of buttons to the one in the common lib.

PR #538, opened 2021-07-16
Remove after date filter

elayabharath:


marianamarasoiu (approved):
thanks EB, LGTM!

Can you also remove it from [katikati_common_lib/url_view](https://github.com/larksystems/katikati_common_lib/blob/main/ui/lib/components/url_view/url_view.dart) when you have a minute? thanks!

PR #547, opened 2021-08-12
remove tag group references

danrubel:
This removes references to the deprecated `Tag.group` field.

I've tested this to the best of my ability in an attempt to ensure no exceptions and no functional changes,
but please review with a critical eye.


marianamarasoiu (approved):
yup - `group` is deprecated, and I think we transitioned to `groups` fully a while back. thanks a lot @danrubel , LGTM!

PR #548, opened 2021-08-12
update tag listener method call

danrubel:
This updates the call to the `Tag` listen method
and should land at the same time as https://github.com/larksystems/katikati_common_lib/pull/75.


marianamarasoiu (approved):
LGTM

PR #559, opened 2021-09-02
Various improvements to mandatory filtering

marianamarasoiu:
This fixes some issues with the mandatory filtering:
* live reaction to changes to the filters in Firebase and updates them as needed - fixes https://github.com/larksystems/Katikati-Core/issues/615
* correctly supports mandatory filtering in firebase becoming more permissive

It also adjusts mandatory filtering so that
* only tags added manually to the filter are highlighted on the conversation (likely fit to the use case of using filters to search for messages)
* only the manually added filters are added to the URL when the URL is updated (this is to ensure that when sharing URLs, only the manually added tags are carried over)


marianamarasoiu:
TBR @lukechurch @elayabharath

PR #562, opened 2021-09-06
Show date separators in conversation messages list

elayabharath:
References https://github.com/larksystems/Katikati-Core/issues/526

### DIFF
--- a/webapp/lib/app/nook/view.dart
+++ b/webapp/lib/app/nook/view.dart
@@ -294,7 +294,7 @@ class ConversationPanelView with AutomaticSuggestionIndicator {
     _messages.append(message.renderElement);
     _messageViews.add(message);
     _messageViewsMap[message.messageId] = message;
-    message._message.scrollIntoView();
+    message.renderElement.scrollIntoView();
   }
 
   void padOrTrimMessageViews(int count) {
@@ -309,7 +309,7 @@ class ConversationPanelView with AutomaticSuggestionIndicator {
 
     for (int i = _messageViews.length; i < count; i++) {
       MessageView message = new MessageView('', DateTime.now(), '', '$i');
-      _messages.append(message._message);
+      _messages.append(message.renderElement);
       _messageViews.add(message);
     }
   }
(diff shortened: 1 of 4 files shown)


webapp/lib/app/nook/view.dart
```diff
@@ -413,34 +426,86 @@ class ConversationPanelView with AutomaticSuggestionIndicator {
...
+    renderElement.setAttribute('hidden', 'true');
+  }
+
+  void show() {
+    renderElement.removeAttribute('hidden');
+  }
+
+  String _getDaysAgoSinceToday(DateTime dateTime) {
+    final today = DateTime.now();
+    final difference = today.difference(dateTime);
```
marianamarasoiu, line 455: To make the difference in days work as expected, we need to first remove the number of hours/minutes etc., from both datetimes, otherwise we would get strange behaviour like all messages sent yesterday evening will appear as sent today, if it's only morning (ie. difference between datetimes is < 24h, but across 2 weekdays).

marianamarasoiu (approved):
hi @elayabharath - LGTM with a few nits and comments, thank you!

webapp/lib/app/nook/view.dart
```diff
@@ -413,34 +426,93 @@ class ConversationPanelView with AutomaticSuggestionIndicator {
...
+
+    if (differenceDays <= 1) {
+      return "Today";
+    } else if (differenceDays <= 2) {
+      return "Yesterday";
+    } else if (differenceDays <= 7) {
+      return "This week";
+    } else if (differenceDays > 7) {
+      return formatter.format(dateTime);
+    }
```
marianamarasoiu, line 474: the week-related calculation isn't working right, because the start of a week is relative to the current day, rather than just being 7 days ago. What we'd need to do is to compute the week number as per [this stackoverflow answer](https://stackoverflow.com/a/54129275), and then compare if they're equal. We can land this PR without it for now, and add this functionality in a follow up, whichever you prefer.

below is the code a bit simplified & comparisons adjusted so they don't need the `+1` from the previous line.
```suggestion
    if (differenceDays < 1) return "Today";
    if (differenceDays == 1) return "Yesterday";
    return formatter.format(dateTime);
```

marianamarasoiu (approved):
thanks @elayabharath  - just a couple more comments on the get-days-ago logic, but otherwise LGTM!

PR #564, opened 2021-09-11
Ask for confirmation before the user leaves the configure pages

elayabharath:
+ Tested by editing the message, not saving, going to a different page (or refresh), alert pops up
+ Then clicked saved, navigate to different page, alert doesn't pop.

Closes https://github.com/larksystems/Katikati-Core/issues/631

### DIFF
--- a/webapp/lib/app/configurator/view.dart
+++ b/webapp/lib/app/configurator/view.dart
@@ -24,6 +24,24 @@ class ConfigurationPageView extends PageView {
   ButtonElement saveConfigurationButton;
   SpanElement saveStatusElement;
 
+  bool _unsavedChanges = false;
+
+  bool get unsavedChanges => _unsavedChanges;
+  set unsavedChanges(bool status) {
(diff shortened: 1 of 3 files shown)


webapp/lib/app/configurator/view.dart
```diff
@@ -24,6 +24,24 @@ class ConfigurationPageView extends PageView {
...
+  bool _unsavedChanges = false;
+
+  bool get unsavedChanges => _unsavedChanges;
+  set unsavedChanges(bool status) {
+    if (status == _unsavedChanges) {
+      return;
+    }
+
+    _unsavedChanges = status;
+    if (status) {
```
marianamarasoiu, line 36: I think this makes it slightly easier to interpret what the if is about:
```suggestion
    if (_unsavedChanges) {
```

marianamarasoiu (approved):
this is great, thanks @elayabharath!  LGTM with a small suggestion

PR #565, opened 2021-09-12
Prominent orange save button on the top right of the config page

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/533


marianamarasoiu (approved):
LGTM, thanks @elayabharath !

PR #569, opened 2021-09-14
Fix the sample messages tooltip disappearing when mouse leaving on tag

elayabharath:
+ When the user leaves the tag, we start a timer for 100ms and by the end of it, if the user is mouse over on the tooltip, we make it stay. Not sure if this is a good strategy, but without embedding this tooltip inside the tag - I can't make the tooltip stay using the mouse enter / leave.

Closes https://github.com/larksystems/Katikati-Core/issues/632

### DIFF
--- a/webapp/lib/app/configurator/tags/view.dart
+++ b/webapp/lib/app/configurator/tags/view.dart
@@ -1,5 +1,6 @@
 library view;
 
+import 'dart:async';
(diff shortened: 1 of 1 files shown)


marianamarasoiu (approved):
Thanks for this @elayabharath - the solution seems to work quite well, and is not too complicated, so LGTM!

PR #570, opened 2021-09-14
Support conversations with no messages

marianamarasoiu:
Needs to be merged after pubspec has been updated with https://github.com/larksystems/katikati_common_lib/pull/84


marianamarasoiu:
TBR

PR #571, opened 2021-09-15
Link messages to respective conversation

elayabharath:
+ Added a default loading state, placeholder text when no messages are present
+ Added a link icon to the text message

Closes https://github.com/larksystems/Katikati-Core/issues/452

### DIFF
--- a/webapp/lib/app/configurator/tags/view.dart
+++ b/webapp/lib/app/configurator/tags/view.dart
@@ -6,6 +6,7 @@ import 'package:dnd/dnd.dart' as dnd;
 import 'package:katikati_ui_lib/components/accordion/accordion.dart';
 import 'package:katikati_ui_lib/components/editable/editable_text.dart';
 import 'package:katikati_ui_lib/components/button/button.dart';
+import 'package:katikati_ui_lib/components/model/model.g.dart';
 import 'package:nook/app/configurator/view.dart';
 export 'package:nook/app/configurator/view.dart';
 import 'package:nook/platform/platform.dart' as platform;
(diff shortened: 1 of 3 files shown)


marianamarasoiu (approved):
Thanks @elayabharath - this is really cool!

Amusingly, this isn't quite addressing the issue you link to! The issue is about showing the link on the sample messages tooltip, and taking the user to the conversation view filtered to that tag (so appending &include-tags="tag-ID to the conversations URL). Your work here is going further by making every message a link, which is even cooler, but if possible I'd like to also have the filtered view link.

So, can you follow this PR up with another one that:
- makes "sample messages for tag" also a link button, that takes the user to the conversations view filtered by that tag
- the link for each message also includes the filtered view for that tag (so the user can easily find other similar messages if they wish)

I've edited the issue to include the above. thanks a lot, and LGTM on this part!

webapp/lib/app/configurator/tags/view.dart
```diff
@@ -6,6 +6,7 @@ import 'package:dnd/dnd.dart' as dnd;
 import 'package:katikati_ui_lib/components/accordion/accordion.dart';
 import 'package:katikati_ui_lib/components/editable/editable_text.dart';
 import 'package:katikati_ui_lib/components/button/button.dart';
+import 'package:katikati_ui_lib/components/model/model.g.dart';
```
marianamarasoiu, line 9: Nit: don't import the generated model directly, but instead import the wrapper `model.dart`
```suggestion
import 'package:katikati_ui_lib/components/model/model.dart';
```

elayabharath:
Ahhh this makes sense - sorry for not understanding it in full. But added them both. Could you take a look at it one more time? Thanks!

webapp/lib/app/configurator/tags/view.dart
```diff
@@ -257,9 +260,16 @@ class SampleMessagesTooltip {
     }
   }
 
-  String _conversationLinkFromMessageID(String messageID) {
-    var conversationID = messageID.replaceAll('nook-message-', '').substring(0, 52);
-    return "/converse/index.html?conversation-id=${conversationID}";
+  String _conversationLinkFromMessageID({String messageId, String tagId}) {
+    Map<String, String> queryParams = {};
+    if(messageId != null) {
```
marianamarasoiu, line 265: nit:
```suggestion
    if (messageId != null) {
```

webapp/lib/app/configurator/tags/view.dart
```diff
@@ -257,9 +260,16 @@ class SampleMessagesTooltip {
...
+      queryParams["conversation-id"] = messageId.replaceAll('nook-message-', '').substring(0, 52);
+    }
+    if(tagId != null) {
```
marianamarasoiu, line 268:
```suggestion
    if (tagId != null) {
```

webapp/lib/app/configurator/tags/view.dart
```diff
@@ -257,9 +260,16 @@ class SampleMessagesTooltip {
     }
   }
 
-  String _conversationLinkFromMessageID(String messageID) {
-    var conversationID = messageID.replaceAll('nook-message-', '').substring(0, 52);
-    return "/converse/index.html?conversation-id=${conversationID}";
+  String _conversationLinkFromMessageID({String messageId, String tagId}) {
```
marianamarasoiu, line 263: I think we need to rename this now, maybe something like: `_linkToFilteredConversationView`?

marianamarasoiu (approved):
you're quick! thanks @elayabharath  - LGTM with a couple of small comments

PR #572, opened 2021-09-15
Add tag UX improvements

elayabharath:
- Adding a new group, selects the title text by default so user can start typing
- The group is expanded so the user can start adding tags

Warning: Do not merge without  https://github.com/larksystems/katikati_common_lib/pull/85

Closes: https://github.com/larksystems/Katikati-Core/issues/575


marianamarasoiu (approved):
LGTM, thanks @elayabharath !

PR #573, opened 2021-09-17
Clicking on tag on tag panel tags the message / conversation

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/623

### DIFF
--- a/webapp/lib/app/nook/view.dart
+++ b/webapp/lib/app/nook/view.dart
@@ -1799,18 +1799,18 @@ class TagActionView implements ActionView {
     action.append(_shortcutElement);
 
     var tagElement = TagView(text, "")
-      ..onSelect = _addTag;
+      ..onSelect = _addTagCommand;
     action.append(tagElement.renderElement);
 
     _buttonElement = new DivElement()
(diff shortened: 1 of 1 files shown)


webapp/lib/app/nook/view.dart
```diff
@@ -1798,17 +1798,22 @@ class TagActionView implements ActionView {
...
     _buttonElement = new DivElement()
       ..classes.add('action__button')
       ..classes.add('action__button--float')
       ..text = buttonText
-      ..onClick.listen((_) => _view.appController.command(UIAction.addTag, new TagData(action.dataset['id'])));
+      ..onClick.listen((_) => _addTag());
     action.append(_buttonElement);
   }
 
+  void _addTag() {
```
marianamarasoiu, line 1813: nit - suggest renaming this to `_addTagCommandCallback`

marianamarasoiu (approved):
LGTM with one small comment, thanks!

webapp/lib/app/nook/view.dart
elayabharath: Fixed!

PR #574, opened 2021-09-17
Make mandatory tags non removable under conversation filters

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/614


marianamarasoiu (approved):

marianamarasoiu:
Can we also add some additional text on hover as per the issue description? something like "This tag cannot be removed from the filter. Please contact your admin if you have any questions". thanks!

elayabharath:
Ahh good point, sorry i missed it earlier! Fixed it by adding a tooltip

PR #575, opened 2021-09-17
On landing on a conversation, the latest message is scrolled into view

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/638

### DIFF
--- a/webapp/lib/app/nook/controller.dart
+++ b/webapp/lib/app/nook/controller.dart
@@ -1395,6 +1395,7 @@ class NookController extends Controller {
     if (conversationsInView.contains(conversation)) {
       // Select the conversation in the list of conversations
       _view.conversationListPanelView.selectConversation(conversation.docId);
+      _view.conversationPanelView.scrollToLatestMessage();
     }
   }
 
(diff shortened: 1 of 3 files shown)


marianamarasoiu (commented):
Sorry EB, this won't work as we'd want. What it will do is scroll to last message both when the conversation is selected (which is what we want), but also when the conversation has been updated (e.g. new message came in, or someone else has tagged a message). This causes problems if for example the user tries to label a message higher up, and the conversation jumps.

Can you find a way to narrow it down so that we scroll to the last message only for that first conversation loading? I believe we already cover the other use cases when a conversation is selected manually.

elayabharath:
Sorry I got carried away for a bit - Move the scroll to bottom function call at the time of "selection" - this is called eventually both when the conversation id query parameter is empty or has initial value.

Could you take a look at it again? Thank you!

marianamarasoiu (commented):
thanks for the iteration EB, but unfortunately this implementation still has the same behaviour as the previous one!

The current implementation would still scroll to the new message, even when the user is looking at an earlier part of the conversation, which would make the conversation jump from under the user.

Can you take a look at https://github.com/larksystems/nook/pull/585 instead, which is trying to only scroll when a new conversation has loaded?

However, I do understand that it's a bit confusing for a new message to arrive and the user not being notified of it, so I've opened https://github.com/larksystems/Katikati-Core/issues/666 to track our fix for this. May I suggest you update this PR to work towards this issue instead? thanks a lot, and sorry to not LGTM this yet!

PR #578, opened 2021-09-17
Select conversation summary to tag, not just active conversation

elayabharath:
+ Show instruction if no conversation tag or message is selected above the tags in the tag panel
+ Deselect a conversation when selecting message, and vice versa

closes https://github.com/larksystems/Katikati-Core/issues/622


marianamarasoiu (approved):
thanks @elayabharath ! just a couple of nits, and also a commit setting the `actionObject` to `null` in a few places instead of `conversation` to support the change of meaning from `conversation` being the default, to `conversation` meaning selected, and `null` being the default

PR #579, opened 2021-09-17
Config for turnline, default tab

elayabharath:
Related to https://github.com/larksystems/Katikati-Core/issues/598

### DIFF
--- a/webapp/lib/app/nook/controller.dart
+++ b/webapp/lib/app/nook/controller.dart
@@ -24,6 +24,9 @@ part 'controller_view_helper.dart';
 
 Logger log = new Logger('controller.dart');
 
+const ENABLE_TURNLINE_PANEL = false;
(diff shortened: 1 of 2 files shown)


webapp/lib/app/nook/controller.dart
```diff
@@ -24,6 +24,9 @@ part 'controller_view_helper.dart';
 
 Logger log = new Logger('controller.dart');
 
+const ENABLE_TURNLINE_PANEL = false;
+const DEFAULT_CONVERSATION_TAB = 'standard_messages';
```
marianamarasoiu, line 28: suggest rename to 
```suggestion
const DEFAULT_PANEL_TAB = 'standard_messages';
```

marianamarasoiu (approved):
great, thanks @elayabharath !

PR #582, opened 2021-09-18
Replace conversation list update with set from scratch

marianamarasoiu:
This fixes the previous logic which wasn't maintaining the sort order of the conversations correctly.

Also tries to reduce jitter in the conversation list when conversations update, but there is more that could be done on this.

And also fixes https://github.com/larksystems/Katikati-Core/issues/510 - this was caused by the lazy list class deciding whether to re-build an item based on `elementOrNull`, which wasn't being set (ie. was always `null`)


elayabharath (approved):
Works well!

PR #584, opened 2021-09-19
Show date separator for conversation list view

elayabharath:
To be merged after: https://github.com/larksystems/katikati_common_lib/pull/87

Closes https://github.com/larksystems/Katikati-Core/issues/526

### DIFF
--- a/webapp/lib/app/nook/view.dart
+++ b/webapp/lib/app/nook/view.dart
@@ -8,6 +8,7 @@ import 'package:katikati_ui_lib/components/accordion/accordion.dart';
 import 'package:katikati_ui_lib/components/tabs/tabs.dart';
 import 'package:katikati_ui_lib/components/url_view/url_view.dart';
 import 'package:katikati_ui_lib/components/snackbar/snackbar.dart';
+import 'package:katikati_ui_lib/utils/dateTime.dart';
 import 'package:katikati_ui_lib/components/nav/button_links.dart';
 import 'package:katikati_ui_lib/components/messages/freetext_message_send.dart';
 import 'package:katikati_ui_lib/components/logger.dart';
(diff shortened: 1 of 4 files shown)


webapp/lib/app/nook/view.dart
```diff
@@ -936,7 +939,19 @@ class ConversationListPanelView {
...
   }
 
-  void updateConversationList(Set<Conversation> conversations) {
+  void onAddConversation(ConversationSummary item) {
+    var currentDateSeparatorString = dateStringForSeparator(item._dateTime);
+    var sortOrder = controller.conversationSortOrder;
+    if (sortOrder == UIConversationSort.alphabeticalById) {
+      item._toggleDateSeparator(false);
+      return;
+    }
```
marianamarasoiu, line 948: I suggest you create a boolean `shouldShowDateSeparator` on the `ConversationListPanelView`, which the controller sets to true or false when the sort order is changed, and here you test.

Otherwise you have part of the logic specific to the sort order in the view, instead of all in one place in the controller. Whether the conversation list will show the date separator or not then becomes part of the API for the UI element, rather than embedding the logic to compute it.

webapp/lib/app/nook/view.dart
```diff
@@ -945,9 +960,9 @@ class ConversationListPanelView {
     for (var conversation in conversations) {
       ConversationSummary summary = _phoneToConversations[conversation.docId];
       if (summary == null) {
-        summary = new ConversationSummary(conversation.docId, "", false);
+        summary = new ConversationSummary(conversation.docId, "", conversation.messages.last.datetime, false);
```
marianamarasoiu, line 963: this needs to include the use case for when there are no messages in the conversation, so something like this: 
```suggestion
        summary = new ConversationSummary(conversation.docId, "", conversation.messages.isEmpty ? null : conversation.messages.last.datetime, false);
```

and can you also just double check if the `ConversationSummary` works ok with a `null` datetime?

marianamarasoiu (commented):
this is great, thanks EB! just a few comments for now, and can I ask you to merge the needed PR in kk_common_lib, and then update the pubspec for this one so I can test it before LGTM-ing? thank you!

elayabharath:
Closing in favour of https://github.com/larksystems/nook/pull/590

PR #591, opened 2021-09-25
Disable buttons if no message or conversation summary is selected

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/672


marianamarasoiu (approved):
thanks @elayabharath - LGTM!

PR #592, opened 2021-09-25
Ignore shortcuts on interactive html elements

elayabharath:
+ Instead of stop propagation on every interactive element, we have a generic function that determines if the shortcut is applicable or not. e.g. this handles "Enter" on a button, but when we author a button element we do not necessarily think of adding stopPropogation onKeyDown

+ Removed stopPropogations on makeEditable as they are not needed anymore

Closes https://github.com/larksystems/Katikati-Core/issues/670


marianamarasoiu (approved):
cool fix, thanks!!

PR #593, opened 2021-09-25
Disable deletion of tags if it is used across any message

elayabharath:
+ Added "cache" for the messages to prevent fetching them multiple times. Please let me know if this is not desirable (since rest of the app is realtime)

Closes https://github.com/larksystems/Katikati-Core/issues/668


marianamarasoiu (approved):
thanks EB - as you anticipated we don't want message cache if possible, since we want the tag list to change following new messages being tagged. Otherwise, LGTM, thank you!

PR #596, opened 2021-09-25
Reflect status in conversation list when user sends messages

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/661


marianamarasoiu (approved):
LGTM, thanks!

PR #597, opened 2021-09-25
Fix conversation width overflow issue

elayabharath:
Merge after https://github.com/larksystems/katikati_common_lib/pull/95


marianamarasoiu (approved):
cool, thanks for the fix @elayabharath !

PR #601, opened 2021-09-26
Add new conversation modal from common lib

elayabharath:


webapp/lib/app/nook/view.dart
```diff
@@ -977,6 +981,12 @@ class ConversationListPanelView {
     panelHeader.append(SpanElement()..className = 'far fa-comments');
     panelHeader.append(_conversationPanelTitle);
 
+    if (ENABLE_NEW_CONVERSTION) {
+      _newConversationModal = NewConversationModal()
+        ..onSubmit = _addNewConversation;
```
marianamarasoiu, line 986: nit:
```suggestion
        ..onSubmit = (List<NewConversationFormData> conversations) => _view.appController.command(UIAction.addNewConversations, NewConversationsData(conversations));
```

webapp/lib/app/nook/view.dart
```diff
@@ -1080,6 +1090,14 @@ class ConversationListPanelView {
     conversationNeedsReply(conversation) ? summary._markUnread() : summary._markRead();
   }
 
+  void _addNewConversation(List<NewConversationFormData> conversations) {
+    _view.appController.command(UIAction.addNewConversations, NewConversationsData(conversations));
+  }
+
```
marianamarasoiu, line 1096: nit:
```suggestion
```

marianamarasoiu (approved):
thanks @elayabharath  - LGTM with a nit - since it's just a command call, it's a bit more straighforward to just write it inline rather than wrap it in its own method.

PR #603, opened 2021-09-30
Fix user presence indicator not being on the conversation item

marianamarasoiu:
Fixes https://github.com/larksystems/Katikati-Core/issues/692

The problem was that the conversation item is now wrapped in a classless div, and the other user presence indicator requires its parent to be relative so it can position itself floating on top, relative to top/bottom.

Also moves it to the bottom so it doesn't appear on top of the datetime


marianamarasoiu:
TBR

PR #607, opened 2021-10-02
Enhancements when adding standard messages under configure page

elayabharath:
+ On adding a category or a message group, we expand the accordion and focus the title to edit
+ minor: changed the double negation under tag view helper

Closes https://github.com/larksystems/Katikati-Core/issues/691


marianamarasoiu (approved):
thanks. LGTM!

PR #608, opened 2021-10-02
Disable tagging when no message is selected and tag group is changed in tag panel

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/689


marianamarasoiu (approved):
LGTM!

PR #609, opened 2021-10-03
Add a "start of conversation" before the first message

elayabharath:
When there are no messages, the conversation panel is empty. The user might be confused if the messages loaded or not.

### DIFF
--- a/webapp/lib/app/nook/view.dart
+++ b/webapp/lib/app/nook/view.dart
@@ -420,6 +420,11 @@ class ConversationPanelView with AutomaticSuggestionIndicator {
     for (int i = 0; i < messagesNo; i++) {
       _messages.firstChild.remove();
     }
+
+    var startOfConversation = DivElement()
+      ..classes.add("messages-date-separator")
+      ..innerText = "You have reached the start of conversation";
+    _messages.append(startOfConversation);
   }
 
   void updateDateSeparators() {
--- a/webapp/web/converse/styles.css
+++ b/webapp/web/converse/styles.css
@@ -345,6 +345,7 @@ main {
     padding: 2px 4px;
     font-size: 0.8em;
     border-radius: 2px;
+    text-align: center;
 }
 
 .notes-panel {


webapp/lib/app/nook/view.dart
```diff
@@ -420,6 +420,11 @@ class ConversationPanelView with AutomaticSuggestionIndicator {
     for (int i = 0; i < messagesNo; i++) {
       _messages.firstChild.remove();
     }
+
+    var startOfConversation = DivElement()
+      ..classes.add("messages-date-separator")
+      ..innerText = "You have reached the start of conversation";
+    _messages.append(startOfConversation);
```
marianamarasoiu, line 427: hey @elayabharath I did some investigation into the issue you flagged (that on the conversation at which the UI opens, this indicator disappears. It seems to be caused by appending it directly to `_messages`, which the rest of this code expects to contain only messages, and not other information. In particular, `_messages.children` gets indexed in `updateMessage` and the messages get updated in place. This means that when the conversations gets initialised as empty, and then the data from firebase comes in, the first message in the conversation will override this div element. It also might mean (I haven't tested it) that when we get updates to a conversation from Firebase, the index would be wrong, and so the wrong message in the UI would get updated.

One quick fix would be on line 363 to have `_messages.children[index + 1] = message.renderElement;` instead of `_messages.children[index] = message.renderElement;` (so taking into account that this element is always the first one). However, I'm a bit worried that this solution is not matching conceptually what we're trying to do with the UI (the fact that the `startOfConversation` element isn't a message), and also that we're introducing an exception (ie. a special case) to the logic. From this perspective I'd be inclined to suggest we add another div element that contains the `startOfConversation` element and the `_messages` element as siblings. I think that will more closely match the conceptual layout with the implementation in the DOM. Can you give this a go and let me know how you get on? thanks!

marianamarasoiu:
Just catching up on some old PRs - closing as implemented in #714

PR #618, opened 2021-10-08
Don't show the conversation if excluded, and hide it when tagged

marianamarasoiu:
Fixes https://github.com/larksystems/Katikati-Core/issues/613


marianamarasoiu:
TBR @lukechurch - as discussed offline, will merge this towards updating the deployments, and will incorporate any feedback subsequently, thanks!

PR #620, opened 2021-10-10
Select a conversation / message tag, press delete to remove tag

elayabharath:
+ Handle selection of conversation, messages, tags one at a time

Prereq before we move the delete into the menu https://github.com/larksystems/Katikati-Core/issues/677


marianamarasoiu:
Thanks @elayabharath !

I did a bit of testing on this since it's quite a big change, and have come across an exception that gets thrown when I:
* open a conversation (click on one in the list on the left)
* select a message in that conversation
* open another conversation (click on a different conversation in the list on the left)

the exception that gets thrown is:
```
Uncaught (in promise) Error: Bad state: No element
    at Object.throw_ [as throw] (errors.dart:209)
    at Array.[dartx.singleWhere] (js_array.dart:299)
    at controller$.NookController.new.updateViewForConversation (controller.dart:1522)
    at controller$.NookController.new.command (controller.dart:1189)
    at view.dart:1406
    at _RootZone.runUnaryGuarded (zone.dart:1384)
    at _ControllerSubscription.new.[_sendData] (stream_impl.dart:357)
    at _DelayedData.new.perform (stream_impl.dart:611)
    at _StreamImplEvents.new.handleNext (stream_impl.dart:730)
    at async._AsyncCallbackEntry.new.callback (stream_impl.dart:687)
    at Object._microtaskLoop (schedule_microtask.dart:41)
    at _startMicrotaskLoop (schedule_microtask.dart:50)
    at async_patch.dart:166
```
Can you take a look as to what might be going on? Thanks!

elayabharath:
@marianamarasoiu I have made the changes wrt UIActionObject. Could you please take a look again? Thank you!

webapp/lib/app/nook/controller.dart
```diff
@@ -1023,6 +1036,47 @@ class NookController extends Controller {
...
+        if (messageTagData.tagId == selectedMessageTagId) {
+          actionObjectState = null;
+          selectedMessageTagId = null;
+          selectedTagMessageId = null;
+        }
+        break;
+      case UIAction.selectMessageTag:
+        command(UIAction.deselectConversationSummary, null);
+        command(UIAction.deselectConversationTag, null);
+        command(UIAction.deselectMessage, null);
```
marianamarasoiu, line 1048: In testing it seems that one can "select" multiple tags if they're on different messages in the same conversation (only in the view though, the controller keeps track correctly). I think it's because we should be having here a call to `command(UIAction.deselectMessageTag, null);`  as the line `messageViewWithId().markSelectedTag()` below doesn't actually deselect tags on other messages.

So let's extract these 4 `deselect...` methods into a helper method 'deselectAllConversationElements` and call it at the beginning of all `select...` methods. Then we can also simplify `markTagSelected` as below

```suggestion
        command(UIAction.deselectMessage, null);
        command(UIAction.deselectMessageTag, null);
```

webapp/lib/app/nook/view.dart
```diff
@@ -373,6 +375,16 @@ class ConversationPanelView with AutomaticSuggestionIndicator {
...
   }
 
+  void markTagSelected(String tagId, bool selected) {
+    _tags.children.forEach((Element e) {
+      if (e.dataset["id"] == tagId) {
+        e.classes.toggle("tag--selected", selected);
+      } else {
+        e.classes.toggle("tag--selected", false);
+      }
+    });
```
marianamarasoiu, line 385: As per comment above, and also to match the intention of the method name, I think this can be simplified to:
```suggestion
    _tags.children
        .filterWhere((Element t) => t.dataset["id"] == tagId)
        .forEach((Element t) => t.classes.toggle("tag--selected", selected));
```
(I haven't run this code, so after accepting can you please check that it compiles and works etc.? thanks!)

webapp/lib/app/nook/view.dart
```diff
@@ -640,6 +652,16 @@ class MessageView {
...
   }
 
+  void markSelectedTag(String tagId, bool selected) {
+    _messageTags.children.forEach((e) {
+      if (e.dataset["id"] == tagId) {
+        e.classes.toggle("tag--selected", selected);
+      } else {
+        e.classes.toggle("tag--selected", false);
+      }
+    });
```
marianamarasoiu, line 662: Same suggestion as above, with the same caveat that it needs testing:
```suggestion
    _messageTags.children
        .filterWhere((Element t) => t.dataset["id"] == tagId)
        .forEach((Element t) => t.classes.toggle("tag--selected", selected));
```

webapp/lib/app/nook/controller.dart
```diff
@@ -35,6 +35,7 @@ enum UIActionObject {
   message,
   loadingConversations,
   addTagInline,
+  tag
```
marianamarasoiu, line 38: nit:
```suggestion
  message,
  tag,
  loadingConversations,
  addTagInline,
```

marianamarasoiu (approved):
thanks so much EB - LGTM with a few small comments!

webapp/lib/app/nook/controller.dart
elayabharath: Ah good catch - I tested by selecting only with in the message and hence didn't catch this. Moved this to a separate function and called it across all the places where we select the elements.

webapp/lib/app/nook/view.dart
elayabharath: Nice - updated this and the below. Tested to make sure this works

PR #630, opened 2021-10-30
Show warning when creating duplicate tags

elayabharath:
Merge after https://github.com/larksystems/katikati_common_lib/pull/117

### DIFF
--- a/webapp/lib/app/configurator/tags/controller.dart
+++ b/webapp/lib/app/configurator/tags/controller.dart
@@ -61,19 +61,6 @@ class TagGroupData extends Data {
   }
 }
 
-class RequestRenameTagData extends Data {
-  String text;
-  Function onUnique;
-  Function onDuplicate;
-
-  RequestRenameTagData(this.text, this.onUnique, this.onDuplicate);
-
-  @override
-  String toString() {
-    return 'RequestRenameTagData($text)';
-  }
-}
-
 TagsConfiguratorController _controller;
 TagsConfigurationPageView get _view => _controller.view;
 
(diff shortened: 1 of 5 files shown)


webapp/lib/app/configurator/tags/view.dart
```diff
@@ -157,7 +157,18 @@ class ConfigureTagView extends TagView {
     draggableTag..onDragStart.listen((_) => dragInProgress = true)..onDragEnd.listen((_) => dragInProgress = false);
 
     onEdit = (text) {
-      _view.appController.command(TagsConfigAction.renameTag, new TagData(tagId, text: text));
+      var requestRenameTagData = new RequestRenameTagData(text, () {
```
marianamarasoiu, line 160: I'd rather we didn't define these as callbacks on a `*Data` object, it's not really a pattern we've used before, and I'm not sure is very easy to read.

Can I suggest instead to:
- change this method to how it was, except that the config action will remain `requestRenameTag`
- add another method for the `ConfigureTagView` class called `showDuplicationWarning` that contains the `onDuplicate` callback code
- the controller code will then call `showDuplicationWarning()` on this tag if duplicate, otherwise will call the `command()` with the `renameTag` action

What do you think?

marianamarasoiu (changes requested):
Hey EB - one comment as I think it would be better to follow the same approach we already do for `TagGroupView.requestToDelete()` - let me know what you think, or if there's any reason why we can't do the same here?

The other comment is to not allow 2 tags with the same name, and instead to show the option to move the tag from one group to another - but we've already spoken about this today, so will be looking forward to reviewing it again afterwards, thanks!

webapp/lib/app/configurator/tags/view.dart
elayabharath: Thanks Mariana, I took the pattern from the JS land - but it makes sense to do this since its an all new pattern.

I have updated the code, except one minor change from what you have suggested - to trigger the modal from the group view instead of tagView directly. Please let me know if that is okay

webapp/lib/app/configurator/tags/controller.dart
```diff
@@ -90,6 +91,18 @@ class TagsConfiguratorController extends ConfiguratorController {
           tagData.groupId: [tag]
         }, startEditing: true);
         break;
+      
+      case TagsConfigAction.requestRenameTag:
+        TagData requestRenameTagData = data;
+        var tagText = requestRenameTagData.text.toLowerCase().trim();
+        var presentTagTexts = tagManager._tags.map((tag) => tag.text.toLowerCase().trim()).toList();
+        
+        if(presentTagTexts.contains(tagText)) {
```
marianamarasoiu, line 100: nit:
```suggestion
        if (presentTagTexts.contains(tagText)) {
```

marianamarasoiu (approved):
This is great, thanks EB!

PR #632, opened 2021-11-21
Show messages time adjusted for local timezone

elayabharath:
Maintain all datetime data in utc, but convert it to local only at the time of display

Fixed a wrong type declaration `{dateTime: DateTime}` - not sure how the compiler hasn't complained about this

Closes https://github.com/larksystems/Katikati-Core/issues/712


marianamarasoiu (approved):
Cool, thanks @elayabharath - LGTM!

PR #633, opened 2021-11-22
Expand all, collapse all buttons for configure tags, messages, conversation filters

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/722

### DIFF
--- a/webapp/lib/app/nook/view.dart
+++ b/webapp/lib/app/nook/view.dart
@@ -1243,7 +1243,20 @@ class ConversationFilter {
     conversationFilter.append(_tagsContainer);
 
     _tagAccordion = Accordion([]);
-    _tagsMenuWrapper.append(_tagAccordion.renderElement);
+    var accordionActions = DivElement()..classes.add('filter-accordion-actions');
+    var expandButton = buttons.Button(buttons.ButtonType.text, buttonText: "Expand all", onClick: (e) {
+      _tagAccordion.items.forEach((AccordionItem item) { item.expand(); });
+    });
+    var collapseButton = buttons.Button(buttons.ButtonType.text, buttonText: "Collapse all", onClick: (e) {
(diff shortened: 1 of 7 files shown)


webapp/lib/app/nook/view.dart
```diff
@@ -1243,7 +1243,20 @@ class ConversationFilter {
     conversationFilter.append(_tagsContainer);
 
     _tagAccordion = Accordion([]);
-    _tagsMenuWrapper.append(_tagAccordion.renderElement);
+    var accordionActions = DivElement()..classes.add('filter-accordion-actions');
+    var expandButton = buttons.Button(buttons.ButtonType.text, buttonText: "Expand all", onClick: (e) {
+      _tagAccordion.items.forEach((AccordionItem item) { item.expand(); });
```
marianamarasoiu, line 1248: nit: 
```suggestion
      _tagAccordion.items.forEach((AccordionItem item) => item.expand());
```
(this is the dart syntax for one liner functions)

webapp/lib/app/nook/view.dart
```diff
@@ -1243,7 +1243,20 @@ class ConversationFilter {
...
+    });
+    var collapseButton = buttons.Button(buttons.ButtonType.text, buttonText: "Collapse all", onClick: (e) {
+      _tagAccordion.items.forEach((AccordionItem item) { item.collapse(); });
```
marianamarasoiu, line 1251:
```suggestion
      _tagAccordion.items.forEach((AccordionItem item) => item.collapse());
```

marianamarasoiu (commented):
just a quick question - why aren't these buttons part of the `Accordion` component and instead we have to create them locally? feels like it should be a generalisable feature?

elayabharath:
Closing in favour of https://github.com/larksystems/katikati_common_lib/pull/120

PR #634, opened 2021-11-26
Added mark as important, copy tag id using menus for config page

elayabharath:
The design implemented slightly deviates from the figma file since
1. there are a lot of hover behaviour in one place
2. menu position on top right make the tag visually cluttered

Merge after https://github.com/larksystems/katikati_common_lib/pull/122


marianamarasoiu (approved):
thanks EB, the approach sounds good to me!

PR #635, opened 2021-11-27
Handle previous / next conversation from conversation summary, shortcuts

elayabharath:
https://github.com/larksystems/Katikati-Core/issues/747

+ Remove 'Enter' shortcut


marianamarasoiu (approved):
Works really nicely, thanks @elayabharath !

PR #640, opened 2021-12-12
Bulk select / deselect conversations using shift key

elayabharath:
Closes https://github.com/larksystems/Katikati-Core/issues/746


marianamarasoiu (approved):
LGTM!

PR #643, opened 2021-12-19
Show unsaved indicator for configure tags

elayabharath:
- editing text, type of tags, move across groups
- rename groups


webapp/lib/app/configurator/tags/controller_view_helper.dart
```diff
@@ -15,20 +15,22 @@ List<MenuItem> getMenuItems(model.Tag tag) {
   ];
 }
 
-void _addTagsToView(Map<String, List<model.Tag>> tagsByCategory, {bool startEditing = false, bool startEditingName = false}) {
+void _addTagsToView(Map<String, List<model.Tag>> tagsByCategory, Set<String> unsavedTagIds, Set<String> unsavedGroupIds, {bool startEditing = false, bool startEditingName = false}) {
```
marianamarasoiu, line 18: as discussed, let's move marking the unsaved tags in a separate function.

Processing unsaved tags & groups here complicates these methods even though it is more efficient performance wise, and it's also conceptually 2 operations - one for adding new tags, the other for updating unsaved status.

webapp/web/configure/tags.css
```diff
@@ -73,3 +73,15 @@
...
+.tag.unsaved::before {
+  font-family: "Font Awesome 5 Free";
+  content: "\f853";
+  display: inline-block;
+  padding-right: 3px;
+  vertical-align: middle;
+  color: var(--orange);
+  font-weight: 900;
+  margin-right: 3px;
+}
```
marianamarasoiu, line 87: nit: github complained about needing a new line at the end of this file
```suggestion
}

```

marianamarasoiu (changes requested):
thanks @elayabharath - as discussed today, this needs a few changes before it's ready, but it's not far off!

marianamarasoiu (changes requested):
thanks for the work on this EB, one small comment and a substantial one - we already keep track of edited & deleted tags in the tag manager, can I ask you to look into whether we can use those variables instead to find out for which tags/groups to show the unsaved indicator? thank you!

webapp/web/configure/tags.css
```diff
@@ -73,3 +73,15 @@
   border: 1px dashed var(--default-border-color);
   background-color: var(--lighter-background-color);
 }
+
+.text-edit.unsaved::before,
+.tag.unsaved::before {
+  font-family: "Font Awesome 5 Free";
+  content: "\f853";
```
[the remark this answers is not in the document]
elayabharath: Ah yes, that makes sense, changed to triangle with exclamation mark

webapp/lib/app/configurator/tags/controller_tag_helper.dart
```diff
@@ -15,6 +15,12 @@ class TagManager {
   Map<String, List<model.Tag>> get tagsByGroup => _tagsByGroup;
   Map<String, List<model.Tag>> _tagsByGroup = {};
 
+  // unsaved tag Ids, group Ids
+  Set<String> _unsavedTagIds = {};
+  Set<String> get unsavedTagIds => _unsavedTagIds;
+  Set<String> _unsavedGroupIds = {};
+  Set<String> get unsavedGroupIds => _unsavedGroupIds;
```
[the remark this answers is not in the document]
elayabharath: I have updated the code to reuse the `editedTags` and `deletedTags`. There are a couple of side effects of this that I noticed though:
1. When a tag is moved across groups, the information in the editedTags doesn't contain the original group it is moved from. Hence solved it by adding additional set to hold the origin groupIds.
2. After a group is renamed, the `editedTags` contains the whole list of tags in that that group. And the UI shows an exclamation mark against all the tags in that group. Though this behaviour is consistent with the data that we hold internally, the action of renaming the group become more dramatic than it really is. Would you have any pointers on how to derive this information from the `editedTags`?

webapp/lib/app/configurator/tags/controller_tag_helper.dart
```diff
@@ -202,6 +202,20 @@ class TagManager {
...
+  Set<String> movedFromGroupIds = {};
+
+  /// Getters for unsaved tag Ids, group Ids derived from editedTags, deletedTags
+  Set<String> get unsavedTagIds => Set.from(List.from(editedTags.keys)..addAll(deletedTags.keys));
+  Set<String> get unsavedGroupIds {
+    window.console.error(movedFromGroupIds);
+    var editedTagGroups = editedTags.values.map((tag) => tag.groups).expand((e) => e); //.toList();
+    var deletedTagGroups = deletedTags.values.map((tag) => tag.groups).expand((e) => e); //.toList();
+    Set<String> setString = Set.from(List.from(editedTagGroups)..addAll(deletedTagGroups))..addAll(movedFromGroupIds);
+    window.console.error(setString);
```
marianamarasoiu, line 215: I'm guessing this is a debug statement and should be removed?
```suggestion
```

marianamarasoiu (approved):
thanks EB - LGTM with just one nit.

As discussed today, this approach is good for now - the long term solution involves adding IDs to the tag group as well as their name, which means we can identify tag group renames more sensibly, but we'll do that in a future PR after the model has been updated to support group IDs.

PR #647, opened 2021-12-27
Collapse / expand the filters bar

elayabharath:


webapp/lib/app/nook/view.dart
```diff
@@ -1003,21 +1005,47 @@ class ConversationListPanelView {
...
       ..classes.add('conversation-list-filters');
-    conversationListPanel.append(panelFilters);
+    conversationListPanel.append(_panelFilters);
 
-    conversationIdFilter = new ConversationIdFilter();
-    panelFilters.append(conversationIdFilter.conversationFilter);
+    var collapsePanelFilters = new DivElement()
+      ..append(SpanElement()..className = "fas fa-chevron-circle-down")
+      ..className = "conversation-filters-toggle"
+      ..onClick.listen((_) { _view.appController.command(UIAction.toggleFiltersPanel, ToggleFiltersPanelData(false)); });
```
marianamarasoiu, line 1015: nit - use syntax for 1-liner function:
```suggestion
      ..onClick.listen((_) => _view.appController.command(UIAction.toggleFiltersPanel, ToggleFiltersPanelData(false)));
```

webapp/lib/app/nook/view.dart
```diff
@@ -1003,21 +1005,47 @@ class ConversationListPanelView {
...
+  }
+
+  void toggleFiltersMenu(bool show) {
+    if (!show) {
+      _panelFilters.hidden = true;
+      _expandPanelFilters.hidden = false;
+    } else {
+      _panelFilters.hidden = false;
+      _expandPanelFilters.hidden = true;
+    }
```
marianamarasoiu, line 1048: nit - a more concise version:
```suggestion
    _panelFilters.hidden = !show;
    _expandPanelFilters.hidden = show;
```

webapp/lib/app/nook/view.dart
```diff
@@ -1003,21 +1005,47 @@ class ConversationListPanelView {
...
+    conversationListPanel.append(_panelFilters);
 
-    conversationIdFilter = new ConversationIdFilter();
-    panelFilters.append(conversationIdFilter.conversationFilter);
+    var collapsePanelFilters = new DivElement()
+      ..append(SpanElement()..className = "fas fa-chevron-circle-down")
+      ..className = "conversation-filters-toggle"
+      ..onClick.listen((_) { _view.appController.command(UIAction.toggleFiltersPanel, ToggleFiltersPanelData(false)); });
+
+    conversationIdFilter = new ConversationIdFilter(collapsePanelFilters);
```
marianamarasoiu, line 1017: Instead of appending the collapse button to the conversation ID filter which just happens to be the first one, can we append it instead to the `_panelFilters` element?

marianamarasoiu (approved):
Thanks EB - a couple of small suggestions, and a comment for you to address, otherwise LGTM!

webapp/lib/app/nook/view.dart
elayabharath: From my understanding, this might be entirely possible since the `_panelFilters` needs to be positioned `relative` which affects the floating filters menu to be rendered beyond the `_panelFilters` (since it is the closest relative positioned element). could you take a look one more time to see if I'm missing something? @marianamarasoiu 

So went ahead and moved this entirely into the conversation ID filter for now.

marianamarasoiu: thanks @elayabharath you're right, it does seem that positioning and z-index interfere with each other, so this wouldn't work.. I'm afraid moving it into the conversation ID filter doesn't seem like a better solution either, as it separates the logic of the collapsing/expanding from the parent element (where it should conceptually be) to a sub-element.

Can I suggest instead that we move the toggle element instantiation it back up to the panel view, and either:
- append it to `conversationIdFilter.conversationFilter` there with a comment as to why we're doing this; or
- wrap the absolute-positioned element into a relative-positioned element, and append that relative element as the first child to `_panelFilters`?

I have a slight preference for the second option as it creates a more straightforward DOM tree, but if it also doesn't work, then we can go with the first.

PR #650, opened 2022-01-07
Add landing page

marianamarasoiu:
> This PR merges into a separate master-multi-tenancy branch, as supporting both single and multi tenancy setup will not work. Once we're ready to transition to multi-tenancy, this branch will be merged into master.

In this PR - at the moment, only the two main pages (landing and project homepage) can interpret correctly the project from the URL).

Am working to add that same capability to the conversation & configure pages etc. but it might need to come in a separate PR.  In the meantime, any additional testing that can be done would be appreciated!


marianamarasoiu:
TBR

PR #655, opened 2022-01-22
Respect sorted order for categories, groups, messages

elayabharath:


marianamarasoiu (approved):
thanks EB, looks good!

PR #658, opened 2022-01-23
Append the message group at the end when adding from the UI

elayabharath:


marianamarasoiu (approved):
this is great, thanks @elayabharath !

PR #660, opened 2022-01-23
Check for actual message change before modifying a message text / translation

elayabharath:


webapp/lib/app/configurator/standard_messages/controller_messages_helper.dart
```diff
@@ -110,6 +110,10 @@ class StandardMessagesManager {
 
   model.SuggestedReply modifyMessage(String messageId, String text, String translation) {
     var message = standardMessages.singleWhere((element) => element.suggestedReplyId == messageId);
+    if ((message.text != null && message.text == text) || (message.translation != null && message.translation == translation)) {
```
marianamarasoiu, line 113: Should this be an AND check?
```suggestion
    if ((message.text != null && message.text == text) && (message.translation != null && message.translation == translation)) {
```

marianamarasoiu (approved):
Good fix! just a question, otherwise LGTM!

webapp/lib/app/configurator/standard_messages/controller_messages_helper.dart
marianamarasoiu: Actually I think this should work as intended and is even shorter:
```suggestion
    if (message.text == text && message.translation == translation) {
```

unless there are times when `message.text` is intentionally `null`?

PR #665, opened 2022-02-03
heartbeat send version

danrubel:
send client version to server during heartbeat


marianamarasoiu (approved):
this is great, thanks @danrubel !

PR #673, opened 2022-02-13
Log IDs on click when developer mode is active

elayabharath:


webapp/lib/app/nook/view.dart
```diff
@@ -594,6 +605,8 @@ class MessageView {
       ..append(_message);
 
     _messageBubble = new DivElement()
+      ..dataset['conversationId'] = conversationId
+      ..dataset['messageId'] = messageId
```
marianamarasoiu, line 609: nit - we can use the same format as the others:
```suggestion
      ..dataset['conversation-id'] = conversationId
      ..dataset['message-id'] = messageId
```

webapp/lib/app/nook/view.dart
```diff
@@ -611,6 +624,8 @@ class MessageView {
     _messageBubble.append(_messageDateTime);
 
     _messageText = new DivElement()
+      ..dataset['conversationId'] = conversationId
+      ..dataset['messageId'] = messageId
```
marianamarasoiu, line 628: nit:
```suggestion
      ..dataset['conversation-id'] = conversationId
      ..dataset['message-id'] = messageId
```

marianamarasoiu (approved):
thanks EB - LGTM!

PR #674, opened 2022-02-13
Handle a lot of tags in the filter conversation UI

elayabharath:
Max height + scroll

Closes https://github.com/larksystems/Katikati-Core/issues/805


marianamarasoiu (approved):
LGTM, thank you!

PR #677, opened 2022-02-20
Populate last inbound turn tags filter menu

elayabharath:
closes https://github.com/larksystems/Katikati-Core/issues/842


marianamarasoiu (approved):
awesome, thanks, LGTM!

PR #683, opened 2022-03-17
Reset for renames of groups, categories, edits of messages, translation

elayabharath:
Changes from the previous commit are still persisting for some reason, will remove it in the following commits

### DIFF
--- a/webapp/lib/app/configurator/standard_messages/controller.dart
+++ b/webapp/lib/app/configurator/standard_messages/controller.dart
@@ -202,8 +202,7 @@ class MessagesConfiguratorController extends ConfiguratorController {
       await Future.wait([platform.updateSuggestedReplies(diffData.editedMessages), platform.deleteSuggestedReplies(diffData.deletedMessages)]);
       _view
         ..unsavedChanges = false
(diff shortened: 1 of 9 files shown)


webapp/lib/app/configurator/standard_messages/controller.dart
```diff
@@ -202,8 +238,7 @@ class MessagesConfiguratorController extends ConfiguratorController {
       await Future.wait([platform.updateSuggestedReplies(diffData.editedMessages), platform.deleteSuggestedReplies(diffData.deletedMessages)]);
       _view
         ..unsavedChanges = false
-        ..showSaveStatus('Saved!', autoHide: true)
-        ..clearUnsavedIndicators();
+        ..showSaveStatus('Published for saving!', autoHide: true);
```
marianamarasoiu, line 241: Nit: "published for saving" won't make sense to the user, so I think we need to stick with just "saved" for now even if it's not quite right
```suggestion
        ..showSaveStatus('Saved!', autoHide: true);
```

marianamarasoiu (approved):
thanks EB - looks good, just a small nit, otherwise LGTM!

PR #684, opened 2022-03-20
Fix the crash where index are missing when adding new category/group

elayabharath:
Sometimes when the category/group index is missing (those saved from a different branch), the user cannot add new category/group


marianamarasoiu (approved):
cool, thanks for the fix! LGTM

PR #686, opened 2022-03-21
Fix the standard messages jump when a new message pops up in the active conversation

elayabharath:
Partially fixes https://github.com/larksystems/Katikati-Core/issues/847


marianamarasoiu (approved):
LGTM, thanks!

PR #687, opened 2022-03-21
Null check for conversations with empty messages

elayabharath:
This causes the number of selected in conversations list to be wrong (possibly one of the reasons)

Potentially fixes https://github.com/larksystems/Katikati-Core/issues/851


marianamarasoiu (approved):
ooh interesting - but this makes sense - I was indeed seeing the issue when doing multi-send to conversations with no incoming messages! 

thanks EB - LGTM!

PR #689, opened 2022-03-27
Show + button for messages tagging with liberal hover area

elayabharath:


marianamarasoiu (approved):
cool, thanks @elayabharath !

marianamarasoiu (approved):
Actually I just got the chance to test this - can I ask for a few changes?

* can we make the + a light grey (same colour as the tag border) - black stands out too much and is distracting if the person is just trying to read the conversation
* can we make it so that the colour of the plus goes darker on hover rather than the background?

Thanks in advance!

PR #692, opened 2022-03-31
Fix for bugs in standard messages page

elayabharath:
+ Do not show reset button as soon as a message is added for the first time
+ Correct for lastIndex of messages, groups, categories that are greater than the parent length
+ Fix the case where the messages in storage might be null hindering comparison i.e. when a new message/group/category is just added

Related https://github.com/larksystems/Katikati-Core/issues/881

### DIFF
--- a/webapp/lib/app/configurator/standard_messages/controller_messages_helper.dart
+++ b/webapp/lib/app/configurator/standard_messages/controller_messages_helper.dart
@@ -30,7 +30,8 @@ class StandardMessagesManager {
       var currentIndex = category.categoryIndex ?? 0;
       return maxIndex > currentIndex ? maxIndex : currentIndex;
     });
-    return lastIndex + 1;
+    var length = localCategories.values.length;
+    return lastIndex >= length ? length : lastIndex + 1;
   }
 
   int getNextGroupIndexInCategory(String categoryId) {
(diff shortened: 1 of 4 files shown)


webapp/lib/app/configurator/standard_messages/controller_messages_helper.dart
```diff
@@ -30,15 +30,17 @@ class StandardMessagesManager {
       var currentIndex = category.categoryIndex ?? 0;
       return maxIndex > currentIndex ? maxIndex : currentIndex;
     });
-    return lastIndex + 1;
+    var length = localCategories.values.length;
+    return lastIndex >= length ? length : lastIndex + 1;
```
marianamarasoiu, line 34: is the below equivalent? if so, let's import the `math` package and use `min()` - I think it's making the logic easier to read:
```suggestion
    return min(localCategories.values.length, lastIndex + 1);
```

(if this works, can you apply it to the below lines as well?)

webapp/lib/app/configurator/standard_messages/controller_messages_helper.dart
```diff
@@ -92,11 +95,17 @@ class StandardMessagesManager {
         unsavedGroupIds.add(storageMessage.groupId);
         unsavedCategoryIds.add(storageMessage.categoryId);
       } else if (storageMessage == null) { // message added
-        editedMessages.add(localMessage);
-        unsavedMessageTextIds.add(localMessage.suggestedReplyId);
-        unsavedMessageTranslationIds.add(localMessage.suggestedReplyId);
-        unsavedGroupIds.add(localMessage.groupId);
-        unsavedCategoryIds.add(localMessage.categoryId);
+        if (localMessage.text != "" || localMessage.translation != "") {
```
marianamarasoiu, line 98: the Dart style guide encourages using `.isNotEmpty()` instead of `!=` for these kinds of checks - can you apply it to all the conditions below?

PR #693, opened 2022-04-09
Reorder message categories, groups

elayabharath:
Needs https://github.com/larksystems/katikati_common_lib/pull/148

### DIFF
--- a/webapp/lib/app/configurator/standard_messages/controller.dart
+++ b/webapp/lib/app/configurator/standard_messages/controller.dart
@@ -1,6 +1,7 @@
 library controller;
 
 import 'dart:html';
+import 'package:katikati_ui_lib/components/accordion/accordion.dart';
 import 'package:katikati_ui_lib/components/logger.dart';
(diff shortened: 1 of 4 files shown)


webapp/lib/app/configurator/standard_messages/controller.dart
```diff
@@ -68,6 +70,27 @@ class StandardMessagesCategoryData extends Data {
   }
 }
 
+class StandardMessagesCategoriesReorderData extends Data {
+  List<String> categoryIds;
+  StandardMessagesCategoriesReorderData(this.categoryIds);
```
marianamarasoiu, line 75: see my comment in the other PR - I think we should pass here only the data that describes the reordering action (so id and index), and not the computed list of ids.

and also the logic of reordering those messages should be moved in this PR (in the controller), which should then call various methods for adjusting the accordion (if possible in place)

marianamarasoiu (approved):
thanks EB, LGTM!

PR #700, opened 2022-05-01
User permissions page

elayabharath:
Access via /users/index.html

+ User permissions are displayed in a table (toggles are checkboxes, text & tags are text boxes)
+ Reads from, saves directly to fire store
+ Handles display of values derived from default, along with reset to default
+ Unsaved indicator in the email header

+ Sticky headers, first column. Hover indicators for easier comprehension. Visual grouping for permissions

Limitations
1. Tags input is not smart, for not requires list of comma separated tag ids
2. Conflicts are not handled
3. Toggle all on/off is not available
4. Removed email does not reflect

### DIFF
--- a/webapp/lib/app/users/controller.dart
+++ b/webapp/lib/app/users/controller.dart
@@ -1,0 +1,242 @@
+library controller;
+
+import 'dart:html';
+import 'package:katikati_ui_lib/components/logger.dart';
+import 'package:nook/app/nook/controller.dart';
+import 'package:nook/platform/platform.dart';
+import 'package:katikati_ui_lib/components/model/model.dart' as model;
+import 'view.dart';
+import 'dart:async';
+
+Logger log = new Logger('controller.dart');
+
+enum UsersAction {
+  updateBoolPermission,
+  updateStringPermission,
+  updateStringSetPermission,
+  savePermissions,
+  resetToDefaultPermission
+}
+
+class UpdateBoolPermission extends Data {
+  String email;
+  String permissionKey;
+  bool value;
+
+  UpdateBoolPermission(this.email, this.permissionKey, this.value);
+}
+
+class UpdateStringPermission extends Data {
+  String email;
+  String permissionKey;
+  String value;
+
+  UpdateStringPermission(this.email, this.permissionKey, this.value);
+}
+
+class ResetToDefaultPermission extends Data {
+  String email;
+  String permissionKey;
+
+  ResetToDefaultPermission(this.email, this.permissionKey);
+}
+
+UsersController controller;
+UsersPageView get _view => controller.view;
+
+class UsersController extends Controller {
+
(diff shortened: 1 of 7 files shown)


webapp/lib/app/users/controller.dart
```diff
@@ -0,0 +1,242 @@
...
+      case UsersAction.updateStringSetPermission:
+        var updateData = data as UpdateStringPermission;
+        var values = updateData.value.split(",").map((e) => e.trim()).toList();
+        if (isDefaultPermission(updateData.email)) {
+          var userDataObj = defaultConfig.toData();
+          userDataObj[updateData.permissionKey] = values;
+          defaultConfig = model.UserConfiguration.fromData(userDataObj);
+          for (var email in userConfigs.keys) {
+            if (email != 'default') {
+              var valueToUpdate = userConfigs[email].toData()[updateData.permissionKey] ?? updateData.value.split(",");
```
marianamarasoiu, line 203: should this be:
```suggestion
              var valueToUpdate = userConfigs[email].toData()[updateData.permissionKey] ?? values;
```

webapp/lib/app/users/view.dart
```diff
@@ -0,0 +1,259 @@
...
+import 'package:katikati_ui_lib/datatypes/user.dart';
+import 'package:nook/view.dart';
+
+import 'controller.dart';
+
+Logger log = new Logger('view.dart');
+
+class Permission {
+  String key;
+  String type;
```
marianamarasoiu, line 13: can we make the `type` an enum containing `bool`, `int`, `string` and `set` ? We don't have uses for `int`s yet, but they can be helpful. We can also assume that `set`s will be `Set<String>` - if/when they're not, we'll work out what to do then)

webapp/lib/app/users/controller.dart
```diff
@@ -0,0 +1,242 @@
...
+          userDataObj[updateData.permissionKey] = updateData.value;
+          userConfigs[updateData.email] = model.UserConfiguration.fromData(userDataObj);
+        }
+        editedConfig[updateData.email] = editedConfig[updateData.email] ?? {};
+        editedConfig[updateData.email][updateData.permissionKey] = updateData.value;
+        _view.markAsUnsaved(updateData.email, false);
+        _view.enableSaveButton(editedConfig.isNotEmpty);
+        break;
+
+      case UsersAction.updateStringPermission:
```
marianamarasoiu, line 171: The code for `update<Type>Permission` is almost identical. I'd be inclined to suggest we have just one user action (`updatePermission`) to which we pass the type of permission - what do you think? (and a similar approach for the view - one entry point to update a permission, but potentially there we keep private methods for each type)

marianamarasoiu (approved):
thanks EB - LGTM with some comments - as discussed, feel free to merge the PR as is, and address these and the remaining features in follow up PRs

PR #701, opened 2022-05-13
Display standard messages as accordion with update in place

elayabharath:


webapp/lib/app/nook/controller_view_helper.dart
```diff
@@ -283,18 +310,12 @@ Map<String, List<model.SuggestedReply>> _groupRepliesIntoGroups(List<model.Sugge
...
   for (model.SuggestedReply reply in replies) {
     // TODO (mariana): once we've transitioned to using groups, we can remove the sequence number fix
-    String groupId = controller.currentConfig.suggestedRepliesGroupsEnabled ?
-        reply.groupId ?? reply.seqNumber.toString() :
-        reply.seqNumber.toString();
+    String groupId = reply.groupId ?? '';
     if (!result.containsKey(groupId)) {
       result[groupId] = [];
     }
     result[groupId].add(reply);
```
marianamarasoiu, line 317: I think we can streamline this some more:
```suggestion
    result.putIfAbsent(reply.groupId, () => []);
    result[groupId].add(reply);
```

marianamarasoiu (approved):
LGTM, thanks EB!

PR #703, opened 2022-05-22
Config tag refactor to include local & storage

elayabharath:
+ New tag manager to include local, storage data, simplified APIs
+ View & view helpers uses group id as the identifier instead of group name
+ Promise await store methods for more reliable save indicator

Known bugs:
+ newly created tag not updating on save, duplicates
+ group's unsaved indicator not cleared
+ _removeTagsFromView should not be called on already deleted group
+ group name can be duplicate, need to build checks
+ Storage vs local diffs are yet to be added
+ Error checks on the TagManager methods


marianamarasoiu (approved):
Thanks EB - looks great!

PR #710, opened 2022-08-03
Add multi-send support

marianamarasoiu:


marianamarasoiu:
Thanks @lukechurch for the offline review, I've updated the link text as per your suggestion

